Issue #40, opened 2018-12-11
Failed to load list of eval filenames from data/list.eval / Segmentation fault


aaronk6:
Hi guys,

I’m pretty new to this, so please forgive if I’m missing something obvious.

[I’ve posted to the mailing list because Tesseract sometimes confuses the digit 4 with a 9 in the material I’m currently processing.](https://groups.google.com/d/msg/tesseract-ocr/Z32zbbVXFuk/RfefWYCOCwAJ) Someone over there pointed me to your project.

So what I would like to do now is to finetune the Latin script to fix the recognition errors I’m seeing. If I understand this correctly, I’ll need to go to `data/ground-truth` and create files there for Tesseract to learn from, e.g:

_April_2014.gt.txt_

```
April 2014
```

_April_2014.tif_

![april_2014](https://user-images.githubusercontent.com/446063/49767845-71603280-fcda-11e8-8712-892646b31ab5.png)

Also, I’ve cloned https://github.com/tesseract-ocr/tessdata.

What else do I need to do? The reason I’m asking is because I get an error when executing the following:

```
$ make -j4 training START_MODEL=Latin TESSDATA=/home/vagrant/tessdata/script
python generate_line_box.py -i "data/ground-truth/April_2012.tif" -t "data/ground-truth/April_2012.gt.txt" > "data/ground-truth/April_2012.box"
python generate_line_box.py -i "data/ground-truth/April_2013.tif" -t "data/ground-truth/April_2013.gt.txt" > "data/ground-truth/April_2013.box"
python generate_line_box.py -i "data/ground-truth/April_2014.tif" -t "data/ground-truth/April_2014.gt.txt" > "data/ground-truth/April_2014.box"
wget -Odata/radical-stroke.txt 'https://github.com/tesseract-ocr/langdata_lstm/raw/master/radical-stroke.txt'
--2018-12-11 00:00:10--  https://github.com/tesseract-ocr/langdata_lstm/raw/master/radical-stroke.txt
Resolving github.com (github.com)... 192.30.253.112, 192.30.253.113
Connecting to github.com (github.com)|192.30.253.112|:443... tesseract data/ground-truth/April_2012.tif data/ground-truth/April_2012 --psm 6 lstm.train
tesseract data/ground-truth/April_2014.tif data/ground-truth/April_2014 --psm 6 lstm.train
find data/ground-truth -name '*.box' -exec cat {} \; > "data/all-boxes"
connected.
[...]

PR #44, opened 2018-12-13
Fail early if not enough ground truth data provided, fix #42

### LINKED ISSUE #42: Add exception handling for GT sets which consist of too few files
Cf. https://github.com/OCR-D/ocrd-train/issues/40

Having fewer than 10 lines results in an empty eval list.

kba:


wrznr:
```
Fail if 1 - RATIO * no < 1
```

kba:
> Fail if (1 - RATIO) * no < 1

Implemented in db1669a

wrznr:
@kba So, we are ready to merge here?

README.md
```diff
@@ -99,6 +99,7 @@ Run `make help` to see all the possible targets and variables:
     PSM                Page segmentation mode. Default: 6
     RANDOM_SEED        Random seed for shuffling of the training data. Default: 0
     RATIO_TRAIN        Ratio of train / eval training data. Default: 0.90
+    MIN_GT_FILES       Minimum number of GT files. Default: 10
 ```
```
stweil: This looks like a relict from the original pull request.

Makefile
```diff
@@ -1,6 +1,6 @@
 export
 
-# Make sure that sort always uses the same sort order.
+## Make sure that sort always uses the same sort order.
 LC_ALL := C
```
stweil: I am curious: what do the double `##` signify?

stweil:
I'm afraid my latest commits have created a merge conflict.

stweil:
I just tested the scenario with a very small set of ground truth and came to the conclusion that there is no need to check the size of the set.

We currently have a rounding problem: a set of 9999 results is split in 8999 and 999, so one set member gets lost.

wrznr:
@stweil This PR is a solution for #42 which was triggered by #40. So it seems to be real problem. Do you think we should ignore such edge cases?

stweil:
We could ignore them in the `Makefile` and let the Tesseract code handle them.

Pull request #84 is a new implementation which supersedes #44. It fixes the rounding problem, and I also implemented the corner cases of empty evaluation set and empty training set.

README.md
stweil: @kba, this commit is good but conflicts with my change which only fixed `TESSERACT_VERSION`.

Makefile
kba: [`makefile-parser`](https://github.com/kba/makefile-parser) won't pick it up as a comment on that variable and it won't be listed in the [help target](https://github.com/kba/makefile-parser#make-help).

stweil:
@kba, I added your commit for `README.md` to Git master, so I think this pull request can now be closed.

stweil:
@kba, I think we can delete the branch.

kba:
Sure go ahead

PR #56, opened 2019-02-27
[Do not merge] Use Wordstr box option for images without ground truth

Shreeshrii:
See https://github.com/tesseract-ocr/tesseract/issues/2267 and https://github.com/OCR-D/ocrd-train/issues/52

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ $(LAST_CHECKPOINT): unicharset lists $(PROTO_MODEL)
 	mkdir -p data/checkpoints
 	lstmtraining \
 	  --traineddata $(PROTO_MODEL) \
-      --old_traineddata $(TESSDATA)/$(START_MODEL).traineddata \
+          --old_traineddata $(TESSDATA)/$(START_MODEL).traineddata \
 	  --continue_from data/$(START_MODEL)/$(START_MODEL).lstm \
 	  --model_output data/checkpoints/$(MODEL_NAME) \
 	  --train_listfile data/list.train \
--- a/Makefile
+++ b/Makefile
@@ -101,15 +101,16 @@ data/unicharset: $(ALL_BOXES)
 	merge_unicharsets data/$(START_MODEL)/$(START_MODEL).lstm-unicharset $(GROUND_TRUTH_DIR)/my.unicharset  "$@"
 else
 data/unicharset: $(ALL_BOXES)
-	unicharset_extractor --output_unicharset "$@" --norm_mode 1 "$(ALL_BOXES)"
+	unicharset_extractor --output_unicharset "$@" --norm_mode $(NORM_MODE) "$(ALL_BOXES)"
 endif
 
 $(ALL_BOXES): $(sort $(patsubst %.tif,%.box,$(wildcard $(GROUND_TRUTH_DIR)/*.tif)))
 	find $(GROUND_TRUTH_DIR) -name '*.box' -exec cat {} \; > "$@"
 
-$(GROUND_TRUTH_DIR)/%.box: $(GROUND_TRUTH_DIR)/%.tif $(GROUND_TRUTH_DIR)/%.gt.txt
-	python generate_line_box.py -i "$(GROUND_TRUTH_DIR)/$*.tif" -t "$(GROUND_TRUTH_DIR)/$*.gt.txt" > "$@"
+$(GROUND_TRUTH_DIR)/%.box: $(GROUND_TRUTH_DIR)/%.tif 
+	tesseract "$(GROUND_TRUTH_DIR)/$*.tif" "$(GROUND_TRUTH_DIR)/$*" -l $(START_MODEL) --psm 6 wordstrbox > "$@"
 
+lstmf: $(ALL_LSTMF)
 $(ALL_LSTMF): $(sort $(patsubst %.tif,%.lstmf,$(wildcard $(GROUND_TRUTH_DIR)/*.tif)))
 	find $(GROUND_TRUTH_DIR) -name '*.lstmf' -exec echo {} \; | sort -R -o "$@"
 
@@ -124,6 +125,7 @@ $(PROTO_MODEL): data/unicharset data/radical-stroke.txt
 	  --input_unicharset data/unicharset \
 	  --script_dir data/ \
 	  --output_dir data/ \
+	  --pass_through_recoder \
 	  --lang $(MODEL_NAME)
 
 ifdef START_MODEL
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -101,15 +101,16 @@ data/unicharset: $(ALL_BOXES)
...
+	unicharset_extractor --output_unicharset "$@" --norm_mode $(NORM_MODE) "$(ALL_BOXES)"
 endif
 
 $(ALL_BOXES): $(sort $(patsubst %.tif,%.box,$(wildcard $(GROUND_TRUTH_DIR)/*.tif)))
 	find $(GROUND_TRUTH_DIR) -name '*.box' -exec cat {} \; > "$@"
 
-$(GROUND_TRUTH_DIR)/%.box: $(GROUND_TRUTH_DIR)/%.tif $(GROUND_TRUTH_DIR)/%.gt.txt
-	python generate_line_box.py -i "$(GROUND_TRUTH_DIR)/$*.tif" -t "$(GROUND_TRUTH_DIR)/$*.gt.txt" > "$@"
+$(GROUND_TRUTH_DIR)/%.box: $(GROUND_TRUTH_DIR)/%.tif 
+	tesseract "$(GROUND_TRUTH_DIR)/$*.tif" "$(GROUND_TRUTH_DIR)/$*" -l $(START_MODEL) --psm 6 wordstrbox > "$@"
```
wrznr: This is a pretty new thing, right? Although it would be nice to get rid of `generate_line_box.py`, the dependency on `START_MODEL` is a problem for training from scratch as the Makefile is set up right now. Is the model actually used while generating the box files? If not, we could simply use an empty model or include the box generation in the `ifdef` construction above.

Makefile
```diff
@@ -133,12 +135,10 @@ $(LAST_CHECKPOINT): unicharset lists $(PROTO_MODEL)
 	  --traineddata $(PROTO_MODEL) \
           --old_traineddata $(TESSDATA)/$(START_MODEL).traineddata \
 	  --continue_from data/$(START_MODEL)/$(START_MODEL).lstm \
-	  --net_spec "[1,36,0,1 Ct3,3,16 Mp3,3 Lfys48 Lfx96 Lrx96 Lfx256 O1c`head -n1 data/unicharset`]" \
```
wrznr: Why is it superfluous? It was used to fix the number of output classes (i.e. characters).

Makefile
```diff
@@ -133,12 +135,10 @@ $(LAST_CHECKPOINT): unicharset lists $(PROTO_MODEL)
...
 	  --model_output data/checkpoints/$(MODEL_NAME) \
-	  --learning_rate 20e-4 \
```
wrznr: Is there a reasonable default?

Makefile
wrznr: Also, the parameter file is not installed if I use the (latest) tesseract version form github.

wrznr: It is maybe not even part of the repo:
```shell
kmw@barclay:~/built/tesseract$ find . -type f -name "*struct"
kmw@barclay:~/built/tesseract$ find . -type f -name "*struct*"
./src/cutil/.deps/structures.Plo
./src/cutil/.libs/structures.o
./src/cutil/structures.lo
./src/cutil/structures.o
./src/cutil/structures.cpp
./src/cutil/structures.h
./src/ccstruct/ccstruct.h
./src/ccstruct/libtesseract_ccstruct.la
./src/ccstruct/ccstruct.lo
./src/ccstruct/.deps/ccstruct.Plo
./src/ccstruct/.libs/libtesseract_ccstruct.a
./src/ccstruct/.libs/ccstruct.o
./src/ccstruct/ccstruct.o
./src/ccstruct/ccstruct.cpp
```

Shreeshrii: Thank you for pointing out about the `wordstrbox` config file. I have submitted a PR to fix that. https://github.com/tesseract-ocr/tesseract/pull/2280

Shreeshrii: I foresee additions to the makefile to handle the multiple options:

1. `generate_line_box.py` is useful for cases where the line images and ground truth is available.

2. `wordstrbox` is useful where scanned images are available and an easily editable box file is required for LSTM training. Please note that the makefile should not continue directly after these box files are generated because the box files will not be correct with the text.

The reason that a START_MODEL is required here is that the box files need to be generated with the best possible transcription for the scanned image. An empty START_MODEL by default uses `eng` - so it will not work for other languages.

Shreeshrii: This was not supposed to be a replacement makefile but rather an example of how to use the new wordstrbox option.

Makefile
Shreeshrii: There are 4 different options for LSTM training - see https://github.com/tesseract-ocr/tesseract/wiki/TrainingTesseract-4.00#tutorial-guide-to-lstmtraining

- Training From Scratch
- Fine Tuning for Impact
- Fine Tuning for ± a few characters
- Training Just a Few Layers

Complete  --net_spec is given only for training from scratch. The other three options do not require that.

Makefile
Shreeshrii: This is what has been written by Ray in https://github.com/tesseract-ocr/tesseract/wiki/TrainingTesseract-4.00#training-from-scratch.

Makefile
Shreeshrii:
```
	lstmtraining \
	  --traineddata $(PROTO_MODEL) \
          --old_traineddata $(TESSDATA)/$(START_MODEL).traineddata \
	  --continue_from data/$(START_MODEL)/$(START_MODEL).lstm \
	  --model_output data/checkpoints/$(MODEL_NAME) \
	  --train_listfile data/list.train \
	  --eval_listfile data/list.eval \
	  --max_iterations 400	lstmtraining \
	  --traineddata $(PROTO_MODEL) \
          --old_traineddata $(TESSDATA)/$(START_MODEL).traineddata \
	  --continue_from data/$(START_MODEL)/$(START_MODEL).lstm \
	  --model_output data/checkpoints/$(MODEL_NAME) \
	  --train_listfile data/list.train \
	  --eval_listfile data/list.eval \
	  --max_iterations 400
```

The above is meant to be for the option - https://github.com/tesseract-ocr/tesseract/wiki/TrainingTesseract-4.00#fine-tuning-for--a-few-characters

 --max_iterations should be 3600 as per Ray's example.

Makefile
wrznr: I see. The four options for training will then have to be reflected in a larger modification of the Makefile.

wrznr:
@Shreeshrii Many thanks for this PR and your explanations. We will try to come up with a (more) general solution.

Makefile
Shreeshrii: >The four options for training will then have to be reflected in a larger modification of the Makefile.

Yes.

In the tesseract repo, tesstrain.sh is used to create synthetic training data - box/tiff pairs from fonts and training_text. That is the third option, in addition to the two I mentioned above.

Shreeshrii:
Improved version of Makefile provided in https://github.com/OCR-D/ocrd-train/pull/66

Issue #71, opened 2019-08-13
Use reproducible list of files for training and eval


stweil:
Currently the list is created from all `lstmf` files by using `sort -R` which creates a randomly sorted list. That solution gives results which cannot be reproduced which is unwanted for several reasons.

Using a pseudo random sort with a known seed which can optionally be set by the user would fix this.

stweil:
https://github.com/stweil/ocrd-train/commit/32b63c5c28173d7f7ac2cf23da95714e23045190 implements a possible solution.

stweil:
From @wrznr (see #72): 
> There is a way to pass seeds to sort -R:
> seq 1 100 | sort -R --random-source=<(openssl enc -aes-256-ctr -pass pass:"42" -nosalt </dev/zero 2>/dev/null)
where 42 is the seed. This may make the additional script superfluous.

No need for `openssl` as far as I see. This works, too:

    seq 1 100 | sort -R --random-source=<(echo 42)

wrznr:
Even better. Although, python may be faster, I'd like to stick with `coreutils`.

stweil:
It only makes a difference of about 30 s on a large data set with 300000 lines, so sticking to coreutils is reasonable. Done in PR #76.

stweil:
There remains a problem:

The code currently shuffles the result lines from `find`. Those lines are unsorted. So they must be sorted before they are shuffled to get a reproducible result.

stweil:
Pull request #77 addresses that additional problem.

stweil:
> This works, too

The simple code without `openssl` works on MacOS (where I had run that test), but fails on Linux. I added a commit to PR #77 to fix that.

stweil:
The lists for training and evaluation are now reproducible. First tests with the latest code show that the training results are nevertheless not reproducible: a comparison of the same training running on one machine with latest Tesseract . This needs more examination. I therefore updated the subject of this issue to cover that.

My current hypothesis is that the different results might be caused by differences in the calculation of the dot product. To test this, the same training must be run with identical Tesseract versions on different machines with similar hardware (AVX support). Maybe Tesseract must also be restricted to a single thread.

wrznr:
Very, very interesting.

Shreeshrii:
You can try with `--sequential_training` to check if that helps.

stweil:
[...]

PR #76, opened 2019-08-15
Reproducible shuffling of training data

stweil:
Add a seed which can be optionally set by the user for sort -R.


stweil:
This fixes issue #71.

PR #77, opened 2019-08-16
Sort training data before shuffling to get reproducible order

stweil:


stweil:
See issue #71.

PR #81, opened 2019-08-29
Makefile: Search files in GROUND_TRUTH_DIR recursively and other small improvements

stweil:


Makefile
```diff
@@ -115,28 +117,28 @@ data/unicharset: $(ALL_BOXES)
 	unicharset_extractor --output_unicharset "$@" --norm_mode 1 "$(ALL_BOXES)"
 endif
 
-$(ALL_BOXES): $(sort $(patsubst %.tif,%.box,$(wildcard $(GROUND_TRUTH_DIR)/*.tif)))
+$(ALL_BOXES): $(patsubst %.tif,%.box,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif'))
```
kba: Won't this reintroduce the issue https://github.com/OCR-D/ocrd-train/commit/bebc7db656aebd92ae3105873a2288724c438b99 tried to fix?

Makefile
```diff
@@ -115,28 +117,28 @@ data/unicharset: $(ALL_BOXES)
...
 	find $(GROUND_TRUTH_DIR) -name '*.box' | xargs cat > "$@"
 
-$(GROUND_TRUTH_DIR)/%.box: $(GROUND_TRUTH_DIR)/%.tif $(GROUND_TRUTH_DIR)/%.gt.txt
-	python3 generate_line_box.py -i "$(GROUND_TRUTH_DIR)/$*.tif" -t "$(GROUND_TRUTH_DIR)/$*.gt.txt" > "$@"
+%.box: %.tif %.gt.txt
+	python3 generate_line_box.py -i "$*.tif" -t "$*.gt.txt" > "$@"
 
-$(ALL_LSTMF): $(sort $(patsubst %.tif,%.lstmf,$(wildcard $(GROUND_TRUTH_DIR)/*.tif)))
+$(ALL_LSTMF): $(patsubst %.tif,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif'))
```
kba: ditto

Makefile
```diff
@@ -115,28 +117,28 @@ data/unicharset: $(ALL_BOXES)
...
 	# https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html#Random-sources
 	find $(GROUND_TRUTH_DIR) -name '*.lstmf' | sort | \
 	  sort -R --random-source=<(openssl enc -aes-256-ctr -pass pass:"$(RANDOM_SEED)" -nosalt </dev/zero 2>/dev/null) > "$@"
 
-$(GROUND_TRUTH_DIR)/%.lstmf: $(GROUND_TRUTH_DIR)/%.box
-	tesseract $(GROUND_TRUTH_DIR)/$*.tif $(GROUND_TRUTH_DIR)/$* --psm $(PSM) lstm.train
+%.lstmf: %.box
+	tesseract $*.tif $* --psm $(PSM) lstm.train
```
kba: Will all make versions expand `%` to the subdir? I think that was why we made that explicit here, because some versions of make apparently did not.

Makefile
```diff
@@ -198,12 +200,9 @@ tesseract.built: tesseract-$(TESSERACT_VERSION)
 	cd $< && \
 		sh autogen.sh && \
 		PKG_CONFIG_PATH="$(LOCAL)/lib/pkgconfig" \
-		LEPTONICA_CFLAGS="-I$(LOCAL)/include/leptonica" \
 			./configure --prefix=$(LOCAL) && \
 		LDFLAGS="-L$(LOCAL)/lib"\
-			make -j$(CORES) && \
-		make install && \
-		make -j$(CORES) training-install && \
+			make -j$(CORES) install training-install && \
```
kba: I'm pretty sure there was a reason to have the `install` target separate from the default target. But cannot remember off the top of my hat. @wrznr?

Makefile
stweil: I don't think so. It works fine with GT4HistOCR (more than 300000 data sets). There is no external shell command with a potentially too long argument list, only Makefile macros are used and these macros should work without arbitrary limits.

Makefile
stweil: See answer above.

Makefile
stweil: I only tested GNU make (standard for most Linux distributions, but also on macOS).

Makefile
stweil: I'm pretty sure that the new code here is fine. :-) If it did not work, I'd consider that a bug in the Tesseract build system and certainly fix it, but I am not aware of such problems and use that kind of build very often.

Separating binary build and install is reasonable when installation needs root rights (sudo), but that's not the case here.

Makefile
stweil: In the old code the [lines 181 and 182](https://github.com/OCR-D/ocrd-train/commit/bebc7db656aebd92ae3105873a2288724c438b99#diff-b67911656ef5d18c4ae36cb6741b7965L181) would obviously fail with GT4HistOCR.

Makefile
kba: > reasonable when installation needs root rights (sudo)

That was probably the reason.

PR #82, opened 2019-08-29
Write more generated files to model specific output directory

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ TESSDATA =  $(LOCAL)/share/tessdata
 # Name of the model to be built. Default: $(MODEL_NAME)
 MODEL_NAME = foo
 
+# Output directory for generated files. Default: $(OUTPUT_DIR)
 OUTPUT_DIR = data/$(MODEL_NAME)
 
 # Name of the model to continue from. Default: '$(START_MODEL)'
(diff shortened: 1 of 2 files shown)


stweil:
This allows running more than one training at the same time.

Makefile
```diff
@@ -11,13 +11,15 @@ TESSDATA =  $(LOCAL)/share/tessdata
 # Name of the model to be built. Default: $(MODEL_NAME)
 MODEL_NAME = foo
 
+OUTPUT_DIR = data/$(MODEL_NAME)
```
kba: Add a comment before and re-run shinclude to have this variable listed in `make` / `make help`?

stweil: Explicitly exposing the output directory to the user would be an additional new feature. I had not planned that in my commit to keep the user interface restricted to things which are really needed, but of course it can be easily added if desired.

stweil:
Explicitly exposing the output directory to the user would be an additional new feature. I had not planned that in my commit to keep the user interface restricted to things which are really needed, but of course it can be easily added if desired.

Makefile
kba: > Explicitly exposing the output directory to the user would be an additional new feature. 

I think it is a new feature and useful and users should be made aware of it.

stweil:
Fixed merge conflict and added documentation manually (no shinclude) now.

PR #83, opened 2019-08-29
Add new parameter NET_SPEC for the network specification

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 export
 
-# Make sure that sort always uses the same sort order.
+## Make sure that sort always uses the same sort order.
 LC_ALL := C
 
 SHELL := /bin/bash
@@ -38,7 +38,7 @@ GROUND_TRUTH_DIR := data/ground-truth
 MAX_ITERATIONS := 10000
 
 # Network specification. Default: $(NET_SPEC)
-NET_SPEC := [1,36,0,1 Ct3,3,16 Mp3,3 Lfys48 Lfx96 Lrx96 Lfx256 O1c1]
+NET_SPEC := [1,36,0,1 Ct3,3,16 Mp3,3 Lfys48 Lfx96 Lrx96 Lfx256 O1c###]
 
 # Normalization Mode - see src/training/language_specific.sh for details. Default: $(NORM_MODE)
 NORM_MODE = 2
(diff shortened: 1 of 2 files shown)


stweil:
@kba, @wrznr, is this modification ready for merging?

Makefile
```diff
@@ -141,14 +145,16 @@ $(PROTO_MODEL): data/unicharset data/radical-stroke.txt
...
+
 ifdef START_MODEL
 $(LAST_CHECKPOINT): unicharset lists $(PROTO_MODEL)
 	mkdir -p data/checkpoints
 	lstmtraining \
 	  --traineddata $(PROTO_MODEL) \
 	  --old_traineddata $(TESSDATA)/$(START_MODEL).traineddata \
 	  --continue_from data/$(START_MODEL)/$(START_MODEL).lstm \
-	  --net_spec "[1,36,0,1 Ct3,3,16 Mp3,3 Lfys48 Lfx96 Lrx96 Lfx256 O1c`head -n1 data/unicharset`]" \
+	  --net_spec "$(subst c1,c`head -n1 data/unicharset`,$(NET_SPEC))" \
```
wrznr: Not completely sure whether this save under all possible configurations (i.e. is it possible that `c1` may be part of the netspec?).

Makefile
```diff
@@ -141,14 +145,16 @@ $(PROTO_MODEL): data/unicharset data/radical-stroke.txt
 	  --output_dir data \
 	  --lang $(MODEL_NAME)
 
+#      19 [1,48,0,1Ct3,3,16Mp3,3Lfys64Lfx96Lrx96Lfx512O1c1]
```
wrznr: Why this comment?

README.md
```diff
@@ -95,6 +95,7 @@ Run `make help` to see all the possible targets and variables:
     TESSDATA_REPO      Tesseract model repo to use. Default: _fast
     GROUND_TRUTH_DIR   Ground truth directory. Default: data/ground-truth
     MAX_ITERATIONS     Max iterations. Default: 10000
+    NET_SPEC           Network specification. Default: [1,36,0,1 Ct3,3,16 Mp3,3 Lfys48 Lfx96 Lrx96 Lfx256 O1c1]
```
wrznr: Not exactly true, `c1` is a variable.

stweil: Would using a well defined placeholder (for example `c###` instead of `c1`) be better? That would also solve your question above.

I had chosen `c1` because that's how Tesseract currently documents the network specifications which were used in the models.

Makefile
stweil: It's a network specifications which was used rather often (19 times) for the Tesseract models. The comment should not be there, so I'll remove it from the PR.

README.md
wrznr: I do not know the net-spec language well enough to judge whether `c1` is really a risk... `c###` is definitely safe.

kba: I'll add a commit that does this.

Makefile
stweil: The comment is removed now.

stweil:
Thank you, @kba. The changes should be ready for merging now.

PR #84, opened 2019-09-03
Check size of ground truth for training and evaluation

stweil:


stweil:
This is based on pull request #82. I'll rebase it as soon as that one was merged.

Issue #85, opened 2019-09-03
Using openssl doesn't work on a Mac, and is overkill on Linux


juliangilbey:
Hi!  Thanks for this training example; it's really helpful!

I found that the Makefile line which sorts the list of training examples breaks on a Mac, presumably because it doesn't have named pipes.  But it is also somewhat of an overkill: sort -R --random-source only reads in 16 bytes from the random source to initialise the sorting, so sort -R is probably adequate here.  shuf reads in much more randomness than this, using the random-source for every step of the permutation, but it will fail on a Mac as the shuf command doesn't exist there.

Best wishes,

Julian

stweil:
Maybe I should reactivate my Python script to sort and shuffle the data. Obviously that would not only be much faster but also improve the portability. @wrznr, what do you think?

kba:
> that would not only be much faster but also improve the portability.

+1

stweil:
@juliangilbey, did you see [my report](https://github.com/tesseract-ocr/tesstrain/wiki/GT4HistOCR) in the Wiki? Maybe it also helps. And of course you are invited to share your results there, too.

juliangilbey:
Thanks - that's a nice report!

I wouldn't go to the effort of using python; I'd just replace

```
find $(GROUND_TRUTH_DIR) -name '*.lstmf' | sort | \
	  sort -R --random-source=<(openssl enc -aes-256-ctr -pass pass:"$(RANDOM_SEED)" -nosalt </dev/zero 2>/dev/null) > "$@"
```

(and I just noticed: there's an extra unnecessary `sort` in that pipeline!) with

```
find $(GROUND_TRUTH_DIR) -name '*.lstmf' | sort -R > "$@"
```

The only difference between this and the original version is that sort -R reads from /dev/urandom if it can, and if not, it does something else reasonably intelligent.  The original version uses the --random-source to get a 16-byte seed, and then uses some form of md5 hashing to generate later randomness.  So it will still be at least as good as the Python version.  (And MacOS supports sort -R.)

Best wishes,

Julian

juliangilbey:
Just a quick clarification: sort -R reads *its seed* from /dev/urandom if it can, and if not, it does something else reasonably intelligent to get a random seed.

stweil:
We want reproducible training results. Therefore we can only use pseudo random sequences. Using `/dev/urandom` is therefore not possible.

juliangilbey:
Ah, I missed that!

[...]

PR #86, opened 2019-09-03
Reproducible shuffling of training data (fix portability issue #85 for macOS)

stweil:
sort -R depends on OpenSSL and does not work on macOS because of missing named pipes,
so replace it with a well defined pseudo random shuffling Python script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,8 +139,7 @@ $(ALL_BOXES): $(patsubst %.tif,%.box,$(shell find $(GROUND_TRUTH_DIR) -name '*.t
 $(ALL_LSTMF): $(patsubst %.tif,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif'))
 	mkdir -p $(OUTPUT_DIR)
 	# https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html#Random-sources
-	find $(GROUND_TRUTH_DIR) -name '*.lstmf' | sort | \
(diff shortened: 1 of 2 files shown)


stweil:
This fixes issue #85.

Makefile
```diff
@@ -139,8 +139,7 @@ $(ALL_BOXES): $(patsubst %.tif,%.box,$(shell find $(GROUND_TRUTH_DIR) -name '*.t
 $(ALL_LSTMF): $(patsubst %.tif,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif'))
 	mkdir -p $(OUTPUT_DIR)
 	# https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html#Random-sources
-	find $(GROUND_TRUTH_DIR) -name '*.lstmf' | sort | \
-	  sort -R --random-source=<(openssl enc -aes-256-ctr -pass pass:"$(RANDOM_SEED)" -nosalt </dev/zero 2>/dev/null) > "$@"
+	find $(GROUND_TRUTH_DIR) -name '*.lstmf' | python3 shuffle.py $(RANDOM_SEED) > "$@"
```
wrznr: No need for Python3 I suppose.

wrznr: I.e. if it works with Python2 as well, we do not have to use Python3 explicitly.

shuffle.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
kba: also could just be `python`

stweil:
We already depend on `python3` for `generate_line_box.py`, so I prefer to keep that uniform.

PR #99, opened 2019-09-19
Create tesseract WordStr box files for given (line) image text pairs

Shreeshrii:
The line images are cropped and then padded with 10 pixel white border and the new bounding box is used for box file. The cropped images work better for training, leading to faster convergence.


zdenop:
thanks.

PR #100, opened 2019-09-19
Create box file for syllables for devanagari script

Shreeshrii:
can be modified to support all Indic scripts by adding appropriate Virama definitions.


zdenop:
thanks

PR #102, opened 2019-10-05
Fixed "UnicodeEncodeError"

mikylucky:
added .encode to char to fix UnicodeEncodeError throwed from example data.

Using provided example data, make training threw "UnicodeEncodeError: 'ascii' codec can't encode character '\xcf' in position 0: ordinal not in range(128)"

### DIFF
--- a/generate_line_box.py
+++ b/generate_line_box.py
@@ -36,9 +36,9 @@ for line in lines:
             char = line[i]
             prev_char = line[i-1]
             if unicodedata.combining(char):
-                print(u"%s %d %d %d %d 0" % ((prev_char + char), 0, 0, width, height))
+                print(u"%s %d %d %d %d 0" % ((prev_char.encode('utf-8') + char.encode('utf-8')), 0, 0, width, height))
             elif not unicodedata.combining(prev_char):
-                print(u"%s %d %d %d %d 0" % (prev_char, 0, 0, width, height))
+                print(u"%s %d %d %d %d 0" % (prev_char.encode('utf-8'), 0, 0, width, height))
         if not unicodedata.combining(line[-1]):
-            print(u"%s %d %d %d %d 0" % (line[-1], 0, 0, width, height))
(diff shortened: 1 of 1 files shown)


stweil:
Could it be that this problem only occurs with Python 3.6, but not with Python 3.7? I recently noticed it, too, but only on Ubuntu Bionic with Python 3.6, not on Debian Buster with Python 3.7.

stweil:
It now no longer throws `UnicodeEncodeError`, but did you have a look on the generated files? In my test, they look like this:

    b'N' 0 0 2074 136 0
    b'e' 0 0 2074 136 0
    b'c' 0 0 2074 136 0

zdenop:
function `encode()` return `bytes` and no string, so IMO this is not correct fix.

mikylucky:
> function `encode()` return `bytes` and no string, so IMO this is not correct fix.

It fixed the issue for me

stweil:
How did your `.box` files look like after that fix?

mikylucky:
> How did your `.box` files look like after that fix?

This is an example box

`b'i' 0 0 976 47 0
b'c' 0 0 976 47 0
b'h' 0 0 976 47 0
b' ' 0 0 976 47 0
b'd' 0 0 976 47 0
b'e' 0 0 976 47 0`

stweil:
Your result:

    b'i' 0 0 976 47 0
    b'c' 0 0 976 47 0
    b'h' 0 0 976 47 0
    b' ' 0 0 976 47 0
    b'd' 0 0 976 47 0
    b'e' 0 0 976 47 0

Correct result:

    i 0 0 976 47 0
    c 0 0 976 47 0
    h 0 0 976 47 0
      0 0 976 47 0
    d 0 0 976 47 0
    e 0 0 976 47 0

stweil:
@mikylucky, I just added a different fix to Git master. Please try that.

mikylucky:
> @mikylucky, I just added a different fix to Git master. Please try that.

Works like a charm!

PR #105, opened 2019-10-06
Sort entries and remove duplicates in all-boxes

stweil:
... and some smaller changes in Makefile.


stweil:
This is one more component to achieve reproducible training.

PR #111, opened 2019-10-11
Normalize text lines for box files and smaller improvements (fix encoding errors)

stweil:


stweil:
I added a script which can be used to normalize ground truth texts.

zdenop:
thanks

PR #114, opened 2019-10-13
Added TESSDATA to README.md

mikylucky:
Added the TESSDATA parameter description to README.md

As per the example at https://github.com/tesseract-ocr/tesstrain/wiki/GT4HistOCR#finetuning-based-on-scriptfraktur

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,7 @@ help:
 	@echo "    LEPTONICA_VERSION  Leptonica version. Default: $(LEPTONICA_VERSION)"
 	@echo "    TESSERACT_VERSION  Tesseract commit. Default: $(TESSERACT_VERSION)"
 	@echo "    TESSDATA_REPO      Tesseract model repo to use. Default: $(TESSDATA_REPO)"
+	@echo "    TESSDATA      	  Path to the .traineddata directory to start finetuning from. Default: $(LOCAL)/share/tessdata
 	@echo "    GROUND_TRUTH_DIR   Ground truth directory. Default: $(GROUND_TRUTH_DIR)"
 	@echo "    OUTPUT_DIR         Output directory for generated files. Default: $(OUTPUT_DIR)"
 	@echo "    MAX_ITERATIONS     Max iterations. Default: $(MAX_ITERATIONS)"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -93,6 +93,7 @@ Run `make help` to see all the possible targets and variables:
     LEPTONICA_VERSION  Leptonica version. Default: 1.78.0
     TESSERACT_VERSION  Tesseract commit. Default: 4.1.0
     TESSDATA_REPO      Tesseract model repo to use. Default: _best
+    TESSDATA           Name of the traineddata file to start finetuning from (without file extension)
```
stweil: The name of the traineddata file is `STARTMODEL.traineddata`.

`TESSDATA` is the name of the directory of the traineddata file.

stweil: If we suggest `TESSDATA` as a variable which may be set by the user, that should also be documented in the `Makefile`.

mikylucky: Done! Let me know if it's ok

Makefile
```diff
@@ -79,6 +79,7 @@ help:
 	@echo "    LEPTONICA_VERSION  Leptonica version. Default: $(LEPTONICA_VERSION)"
 	@echo "    TESSERACT_VERSION  Tesseract commit. Default: $(TESSERACT_VERSION)"
 	@echo "    TESSDATA_REPO      Tesseract model repo to use. Default: $(TESSDATA_REPO)"
+	@echo "    TESSDATA      	  Path to the .traineddata directory to start finetuning from. Default: $(LOCAL)/share/tessdata
```
stweil: Please use space characters here instead of TAB.

Line 9 also still needs a comment now.

README.md
```diff
@@ -93,7 +93,7 @@ Run `make help` to see all the possible targets and variables:
     LEPTONICA_VERSION  Leptonica version. Default: 1.78.0
     TESSERACT_VERSION  Tesseract commit. Default: 4.1.0
     TESSDATA_REPO      Tesseract model repo to use. Default: _best
-    TESSDATA           Name of the traineddata file to start finetuning from (without file extension)
+    TESSDATA           Path to the .traineddata directory to start finetuning from. Default: './usr/share/tessdata'
```
stweil: As the other descriptions don't use apostrophes, I suggest to omit them here, too.

mikylucky:
Thank you for you support!

PR #117, opened 2019-10-17
Support additional line image formats (*.png, *.bin.png, *.nrm.png)

stweil:
Tesseract can handle most common image formats while the Makefile
only had supported *.tif.

Line images from other Open Source OCR software use *.bin.png or *.nrm.png.
The Makefile now handles these, too, so it is no longer necessary to
convert them to *.tif.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,13 +171,14 @@ endif
 # Start training
 training: $(OUTPUT_DIR).traineddata
 
-$(ALL_GT): $(patsubst %.tif,%.gt.txt,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif')) \
-	      $(patsubst %.bin.png,%.gt.txt,$(shell find $(GROUND_TRUTH_DIR) -name '*.bin.png')) \
-	      $(patsubst %.nrm.png,%.gt.txt,$(shell find $(GROUND_TRUTH_DIR) -name '*.nrm.png'))
+$(ALL_GT): $(shell find $(GROUND_TRUTH_DIR) -name '*.gt.txt')
 	@mkdir -p $(OUTPUT_DIR)
 	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs cat | sort | uniq > "$@"
 
 .PRECIOUS: %.box
+%.box: %.png %.gt.txt
+	PYTHONIOENCODING=utf-8 python3 generate_line_box.py -i "$*.png" -t "$*.gt.txt" > "$@"
+
 %.box: %.bin.png %.gt.txt
 	PYTHONIOENCODING=utf-8 python3 generate_line_box.py -i "$*.bin.png" -t "$*.gt.txt" > "$@"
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -130,19 +130,37 @@ $(OUTPUT_DIR)/unicharset: $(ALL_BOXES)
...
-$(ALL_LSTMF): $(patsubst %.tif,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif'))
+$(ALL_LSTMF): $(patsubst %.tif,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.tif')) \
+	      $(patsubst %.bin.png,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.bin.png')) \
+	      $(patsubst %.nrm.png,%.lstmf,$(shell find $(GROUND_TRUTH_DIR) -name '*.nrm.png'))
 	@mkdir -p $(OUTPUT_DIR)
 	find $(GROUND_TRUTH_DIR) -name '*.lstmf' | python3 shuffle.py $(RANDOM_SEED) > "$@"
 
 %.lstmf: %.box
-	tesseract $*.tif $* --psm $(PSM) lstm.train
+	@if test -f "$*.bin.png"; then \
```
wrznr: I am not sure if it is useful to prefer one image extension over another. Couldn't we set the extension via another variable?

stweil: That variable would have to support a list of image name patterns, because at least the mixture of `*.bin.png` and `*.nrm.png` is very common. Maybe we could allow two variables, for example `IMAGE_EXTENSION` (default: `tif`) and `IMAGE_EXTENSION2` (default: undefined or empty).

stweil: Another possibility would be a variable `GT_NAMES` which can be set to a naming scheme, for example `GT_NAMES=ocropus`.

wrznr: But do I interpret your code correctly, if I state that *a mixture of `*.bin.png` and `*.nrm.png`* is not possible given the `if-else` construct you propose here?

stweil: Support for a mixture (like in GT4HistOCR) was a must and triggered these modifications.

That command block is executed for each box file and tries to find the matching image file. It will take the first image name with a corresponding image file.

wrznr: Right. I think a list of extensions (which also defines the precedence) as a variable would be the right way to go. However, I am not sure if this is doable in a Makefile.
```
for ext in extension_list:
    if test -f $*.ext; then
        ...
```
Even dynamic shuffling would be possible to create a real *mixture* of e.g. binary and gray images (remember that ocropy creates both variants if the `--gray` parameter is set).

Shreeshrii:
This should not be closed. 

I suggest that the makefile support `tif`, `png` and the suggested (*.bin.png, *.nrm.png)

wrznr:
@Shreeshrii @stweil Would you agree to have another variable `EXT`?
```shell
(env) $ PYTHONIOENCODING="UTF-8" make -j 4 training MODEL_NAME=htr TESSDATA=/usr/local/share/tessdata/ PSM=13 MAX_ITERATIONS=10000 EXT=.bin.png
```

Shreeshrii:
Please see earlier comments by @stweil. I think his use case has images files with many (two?) different extensions.

wrznr:
@Shreeshrii Right. I forgot. Sorry.

Shreeshrii:
https://stackoverflow.com/questions/41893115/makefile-select-files-by-extension-from-a-variable/41893264 suggests 

apply two filters in one go:

`OBJS = $($(SRCS:%.c=%.o):%.cpp=%.o)`

Can something similar be used for image extensions?

Makefile
kba: Here's a proof-of-concept on how this could be done: https://github.com/OCR-D/docs/blob/master/shuffle-extensions.mk

stweil:
I updated the PR now to fix merge conflicts.

wrznr:
We should have a look at the more generic solution offered by @kba.

stweil:
> We should have a look at the more generic solution offered by @kba.

Up to now I did not manage to get that working.

The current PR at least handles some common cases. Can we apply it as it is and improve it later, for example with a comma separated list of image extensions like `IMAGE_EXTENSIONS=tif,nrm.png,bin.png`.

stweil:
> I suggest that the makefile support tif, png and the suggested (*.bin.png, *.nrm.png)

The latest commit added support for `*.png`, so all those patterns are supported now.

Issue #119, opened 2019-10-25
Feature Request: Script to create Ground Truth from Box files


Shreeshrii:
When using text2image for creating box files, unrenderable words are dropped. Hence there is no guaranteed correspondence between the generated tif and training text. 

So any WordStr box files created for Indic and RTL languages by substituting the text in box file from training text do not work.

It would be helpful to have a script which generates the groundtruth text from the regular text2image generated box files.

I have a bash script which does some sed substitutions to achieve this, but feel that a python script would be a cleaner solution.

Shreeshrii:
```
# uses "λ" for substitution to  get the space delimiters - change for Greek text which uses "λ"
#

gtstring = io.StringIO()
gtfile = open(args.txt, "w", encoding='utf-8')
with io.open(args.box, "r", encoding='utf-8') as boxfile:
        print(''.join(line.replace("  ","λ ").split(' ',1)[0] for line in boxfile if line), file = gtstring)
gt = gtstring.getvalue().replace("λ"," ").replace("\t","\n")
print(gt, file = gtfile)
```

The above works. Is there a better way to do this though?

Shreeshrii:
Tested with the following files - 

```
python3 generate_gt_from_box.py -b tmplines.box.txt -t tmplines.gt.txt
```

[tmplines.box.txt](https://github.com/tesseract-ocr/tesstrain/files/3787738/tmplines.box.txt)
[tmplines.gt.txt](https://github.com/tesseract-ocr/tesstrain/files/3787739/tmplines.gt.txt)

PR #121, opened 2019-11-01
Generate groundtruth from text2image box files

Shreeshrii:
adds feature request in https://github.com/tesseract-ocr/tesstrain/issues/119

### DIFF
--- a/generate_gt_from_box.py
+++ b/generate_gt_from_box.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+
+import argparse
+import io
+import unicodedata
+
+#
+# command line arguments
+#
+arg_parser = argparse.ArgumentParser('''Creates groundtruth files from text2image generated box files''')
+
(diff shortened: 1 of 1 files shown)


generate_gt_from_box.py
```diff
@@ -0,0 +1,30 @@
...
+arg_parser.add_argument('-t', '--txt', nargs='?', metavar='TXT', help='Line text (GT)', required=True)
+
+# Text box file
+arg_parser.add_argument('-b', '--box', nargs='?', metavar='BOX', help='text2image generated box file (BOX)', required=True)
+
+args = arg_parser.parse_args()
+
+#
+# main
+# uses "λ" for substitution to  get the space delimiters - change for Greek text which uses "λ"
```
kba: Why not use a control character outside like `NUL` or an dingbat/emoji instead of the fairly frequent lambda?

PR #125, opened 2019-11-03
Add more documentation for traineddata target

stweil:


stweil:
https://github.com/stweil/tesstrain shows how the result looks like.

PR #127, opened 2019-11-30
Add script to generate reversed RTL text wordstrbox files

Shreeshrii:
Fixes problem for RTL languages with earlier script as reported in https://github.com/tesseract-ocr/tesseract/issues/2669#issuecomment-559106006


Shreeshrii:
[test.gt.txt](https://github.com/tesseract-ocr/tesstrain/files/3906763/test.gt.txt)

test

Shreeshrii:
Currently fonts_dir and fontname are hardcoded in the script. While idelayy it should be a font that can represent 100% of the gt text, it doesn't really matter as we are only interested in the box file.

`--strip_unrenderable_words=false ` should ensure that complete gt is represented in the charbox.

@stweil You can check using a sample from https://github.com/OpenITI/

Shreeshrii:
Please feel free to modify the code to improve/optimize it.

Shreeshrii:
@kba Thank you for your suggestion to use an emoji character for the substitution.  I have done that here.

Shreeshrii:
As per Ray even for RTL languages the processing is done in LTR order. So there is  some code which reverses the input to LTR and at end reverses the output to RTL. 

I have no idea where this is done.  

By using text2image with the groundtruth txt (RTL) the box file is generated in required reverse order (LTR) using a dummy font (since we want to use the scanned image). And then the extracted groundtruth (LTR) is used for a wordstr box file for scanned image.

wrznr:
@Shreeshrii Thanks for your explanations

> processing is done in LTR order

Makes sense. I think that is the default also for `OCRopus` and its successors.

> the box file is generated in required reverse order

Alternatively, we could reverse the text in Python and generate the box files as in `generate_wordstr_box.py`, right? Maybe, we could even reuse the script with a corresponding RTL switch.

Shreeshrii:
That would be great.

Shreeshrii:
>I think that is the default also for OCRopus and its successors.

Do you know if the source data for the Arabic model for OCRopus is available?

wrznr:
Are you referring to a specific model?

Shreeshrii:
Closing as https://github.com/tesseract-ocr/tesstrain/pull/137 provides a cleaner and more elegant solution.

PR #129, opened 2019-12-03
Update description to match actual code

Jertlok:
As the feature for modifying the network layer got removed (85b5ae8), the README file should be updated accordingly.


wrznr:
Thanks @Jertlok.

PR #132, opened 2019-12-10
note pillow dependency

M3ssman:
Added small addition for Python Dependency and explanation when installing tesstrain. Otherwise running into 

```
  File "generate_line_box.py", line 6, in <module>
    from PIL import Image
ModuleNotFoundError: No module named 'PIL'

```
issue when starting to train


wrznr:
I am hesitant to add a `requirements.txt`.
1. This is not a Python project. If we want to make it one, a few other things have to be changed.
2. We do not provide build files or dependency checks for `make` or `tesseract`. Yet, both are necessary to run the training.

M3ssman:
I agree, 
this is not a common Python project and far away from the consistent style of other OCR-D module projects, although it already includes software-management artifacts like a `.pylintrc` at root level (which, in turn, might be a legacy thing).

On the other hand, the targets provided to build both leptonica and tesseract are shaped in usual Makefile-manner, which includes sourcecode-downloads and dependency checks as well and will fail rather verbose if something can't be done as expected - assuming `wget`and regular build-tools are on board.

IMHO, adding a clear marker for Python dependencies increases project transparency and eases the pain to get started. In fact, without Python, i.e. `python3`, and without Pillow, it is impossible to do anything. This functional dependencies must be communicated in a conventional way, which is, in the Python-Universe, a Dependency-File.

wrznr:
@M3ssman Okay. You convinced me: Let's have `requirements.txt`. With respect to your modification of the README, why is having `pip` a must?

README.md
```diff
@@ -24,6 +24,10 @@ autotools (including autotools-archive) and some additional libraries for the
 training tools. See the [installation notes in the tesseract
 repository](https://github.com/tesseract-ocr/tesseract/blob/master/INSTALL.GIT.md).
 
+### python
```
stweil:
```suggestion
### Python
```

README.md
```diff
@@ -24,6 +24,10 @@ autotools (including autotools-archive) and some additional libraries for the
...
+
+You need a recent version of Python 3.x and `pip`. For Image-Processing the Python-Library `Pillow` is used. If you don't have a global Installation, please use the provided Requirements-file `pip install -r requirements.txt`.
```
stweil:
```suggestion
You need a recent version of Python 3.x. For image processing the Python library `Pillow` is used.
If you don't have a global installation, please use the provided requirements file `pip install -r requirements.txt`.
```

M3ssman:
@wrznr Using pip is the common way to install Python packages. To be honest, I personally never used a different tool to manage Python Dependencies. How do you handle this? Do you prefer to clone projects repositories .. ? Well, I guess `pip`is in fact part of a *nix-Python-Installation. The `python-pip`version bundled with official Ubuntu 18.04 LTS Server is outdated (9.x-something instead of current 19.x) and has Issues resolving Dependencies that require wheels, so it is necessary to upgrade `pip` as well _before_ it is really doing what it's supposed to do.

@stweil it's a good point to mention the global installation. When working on shared remote Hosts, with automated CI/CD-Services, it is recommended to use separate `venv` for each project. Therefore, I didn't look that direction.

stweil:
Installing `python-pip` which provides `pip` would be wrong because that is for Python 2.x. Installing `python-pip3` is not needed when using virtual environments because venv typically provides `pip` and `pip3`. That's why I'd not say "You need [...] pip".

requirements.txt
```diff
@@ -0,0 +1 @@
+Pillow==6.2.1
```
kba: `>=` would suffice

wrznr:
@kba Me neither. That's strange.

wrznr:
@M3ssman Could you please modify your PR as @stweil suggests?

stweil:
@wrznr, could you please try once more? I just modified the repository settings. You already were an admin, but maybe that rights were restricted by the team settings.

wrznr:
@stweil Hah: It is because I can not write to https://github.com/ulb-sachsen-anhalt/tesstrain what is perfectly fine. We will have to wait for @M3ssman .

kba:
Can u please make it a >= requirement? Otherwise there might be version issues

kba:
LGTM

M3ssman:
@all Confused me: PR still hangs at `Changes requested`?

kba:
Yeah, but it is ready for merge. @stweil @wrznr feel free I'm afk

wrznr:
I rerequested a review by @stweil. Let's wait until it is done.

stweil:
Thank you, @M3ssman (and also to the reviewers).

Issue #133, opened 2019-12-11
Update documentation on ground truth directory


wrznr:
With https://github.com/tesseract-ocr/tesstrain/commit/0f95a1a7cad11e62df7079fe10b64090ad6a8025 ground truth files have to be placed in `data/$(MODEL_NAME)-ground-truth`. Otherwise `make` won't find them. Update documentation accordingly and delete default `ground-truth` directory.

wrznr:
Fixed via #135

PR #135, opened 2019-12-12
Drop "BUILD_TYPE" and remove default ground truth directory

### LINKED ISSUE #134: What is the idea behind "BUILD_TYPE"?
With https://github.com/tesseract-ocr/tesstrain/commit/480383aeee6f229caecf972b8d4530bcc9584e79 models receive an extra `Plus` when a start model is chosen. @Shreeshrii What is the idea behind this? (I.e. personally, if I set a model name for training I expect my model to have this name after training)

wrznr:
Meanwhile, GT has to be placed in `data/MODEL_NAME-ground_truth`.
This commit fixes the documentation. In addition, the variable
`BUILD_TYPE` is removed to make sure thate the resulting models
are named as requested by the user.

Fixes #134 and #133

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ PUNC_FILE := $(OUTPUT_DIR)/$(MODEL_NAME).punc
 # Name of the model to continue from. Default: '$(START_MODEL)'
 START_MODEL =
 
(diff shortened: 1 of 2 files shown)


stweil:
Thanks. I force-pushed a typo fix for the description and merged the result.

PR #139, opened 2019-12-19
mark box files as PRECIOUS

kba:
... so they aren't treated as intermediate files and deleted on cancel or completion


stweil:
Thanks.

PR #159, opened 2020-04-17
Improvements for RTL

Shreeshrii:

PR #161, opened 2020-04-18
Change Default psm to 13 in Makefile

Shreeshrii:
As per https://github.com/tesseract-ocr/tesstrain/pull/159#issuecomment-615713572


stweil:
Thank you!

PR #171, opened 2020-06-25
generate*.py: raise ValueError if GT is multiline, fix #162

### LINKED ISSUE #162: generate_line_box.py should bomb out if GT has multiple lines
Hi!

Because the generated box files are useless in the case of a multiline image, the current code is very misleading and breaks training if fed multiline ground truths/images.  (I only just realised why my training is not working...)

My suggestion is that the code is modified in two ways:

(1) If len(lines) > 1, then exit with a suitably informative message.

(2) Remove the for loop, and replace "line.strip" with "line[0].strip" in the normalize line.

Best wishes,

   Julian

kba:
`*.gt.txt` should contain a single line of transcribed text.

### DIFF
--- a/generate_gt_from_box.py
+++ b/generate_gt_from_box.py
@@ -2,18 +2,20 @@
 
 import argparse
 import io
-import unicodedata
 
 #
 # command line arguments
 #
-arg_parser = argparse.ArgumentParser('''Creates groundtruth files from text2image generated box files''')
+arg_parser = argparse.ArgumentParser(
+    '''Creates groundtruth files from text2image generated box files''')
 
 # Text ground truth
-arg_parser.add_argument('-t', '--txt', nargs='?', metavar='TXT', help='Line text (GT)', required=True)
+arg_parser.add_argument('-t', '--txt', nargs='?',
+                        metavar='TXT', help='Line text (GT)', required=True)
 
 # Text box file
-arg_parser.add_argument('-b', '--box', nargs='?', metavar='BOX', help='text2image generated box file (BOX)', required=True)
+arg_parser.add_argument('-b', '--box', nargs='?', metavar='BOX',
+                        help='text2image generated box file (BOX)', required=True)
 
 args = arg_parser.parse_args()
 
@@ -25,6 +27,6 @@ args = arg_parser.parse_args()
 gtstring = io.StringIO()
 gtfile = open(args.txt, "w", encoding='utf-8')
 with io.open(args.box, "r", encoding='utf-8') as boxfile:
-        print(''.join(line.replace("  ","λ ").split(' ',1)[0] for line in boxfile if line), file = gtstring)
-gt = gtstring.getvalue().replace("λ"," ").replace("\t","\n")
-print(gt, file = gtfile)
+    print(''.join(line.replace("  ", "λ ").split(' ', 1)[0] for line in boxfile if line), file=gtstring)
+gt = gtstring.getvalue().replace("λ", " ").replace("\t", "\n")
+print(gt, file=gtfile)
--- a/generate_line_box.py
+++ b/generate_line_box.py
@@ -28,17 +28,18 @@ width, height = Image.open(args.image).size
 # load gt
 with io.open(args.txt, "r", encoding='utf-8') as f:
     lines = f.read().strip().split('\n')
+    if len(lines) > 1:
+        raise ValueError("ERROR: %s: Ground truth text file should contain exactly one line, not %s" % (args.txt, len(lines)))
+    line = unicodedata.normalize('NFC', lines[0].strip())
 
-for line in lines:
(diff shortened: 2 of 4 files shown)


kba:
@wrznr @Shreeshrii @stweil Can this be merged?

generate_wordstr_box.py
```diff
@@ -32,11 +32,13 @@
...
-    line = unicodedata.normalize('NFC', line.strip())
-    if line:
-        line = bidi.algorithm.get_display(line)
-        print("WordStr 0 0 %d %d 0 #%s" % (width, height, line))
-        print("\t 0 0 %d %d 0" % (width, height))
+if line:
+    line = unicodedata.normalize('NFC', lines[0].strip())
+    line = bidi.algorithm.get_display(line)
+    print("WordStr 0 0 %d %d 0 #%s" % (width, height, line))
+    print("\t 0 0 %d %d 0" % (width, height))
```
stweil, line 44: The final newline got lost here.

generate_wordstr_box.py
```diff
@@ -32,11 +32,13 @@
...
 # create WordStr line boxes for Indic & RTL
-for line in lines:
-    line = unicodedata.normalize('NFC', line.strip())
-    if line:
-        line = bidi.algorithm.get_display(line)
-        print("WordStr 0 0 %d %d 0 #%s" % (width, height, line))
-        print("\t 0 0 %d %d 0" % (width, height))
+if line:
+    line = unicodedata.normalize('NFC', lines[0].strip())
+    line = bidi.algorithm.get_display(line)
```
stweil, line 42: Moving the conditional `if line:` changes the behaviour. Lines with only whitespace are now handled. I don't think that's what we want here.

kba: oh, right, I resolved that incorrectly. will fix.

generate_wordstr_box.py
```diff
@@ -32,11 +32,13 @@
 # load gt
 with io.open(args.txt, "r", encoding='utf-8') as f:
     lines = f.read().strip().split('\n')
+    if len(lines) > 1:
```
stweil, line 35: As the ground truth text file should contain exactly one line, you might as well check for exactly 1.

kba: 13a58fb

generate_wordstr_box.py
kba: 271d6fa

generate_gt_from_box.py
```diff
@@ -2,18 +2,20 @@
 
 import argparse
 import io
-import unicodedata
```
wrznr, line 5: Why do you propose to remove the import? `unicodedata` is used.

generate_line_box.py
```diff
@@ -28,17 +28,18 @@
...
-            if unicodedata.combining(char):
-                print("%s 0 0 %d %d 0" % ((prev_char + char), width, height))
-            elif not unicodedata.combining(prev_char):
-                print("%s 0 0 %d %d 0" % (prev_char, width, height))
-        if not unicodedata.combining(line[-1]):
-            print("%s 0 0 %d %d 0" % (line[-1], width, height))
-        print("\t 0 0 %d %d 0" % (width, height))
+    if len(lines) == 1:
+        raise ValueError("ERROR: %s: Ground truth text file should contain exactly one line, not %s" % (args.txt, len(lines)))
+    line = unicodedata.normalize('NFC', lines[0].strip())
```
wrznr, line 33: Wow. This is actually a pretty strong modification, right? Are we sure that NFC is the right decision for all the exotic languages for which models are trained for using tesstrain?

generate_gt_from_box.py
kba: Not in this script AFAICS

generate_line_box.py
kba: I didn't introduce the NFC normalization, I just moved the normalization step.

generate_gt_from_box.py
wrznr: Ah, sorry. My bad.

stweil:
See also PR #185.

generate_line_box.py
shaikarshad: Was this value error supposed to be raised for lines greater than one? because the current if condition is checking for length to be 1. I was training on my data whose line length is 1 but getting this issue:

  File "generate_line_box.py", line 32, in <module>
    raise ValueError("ERROR: %s: Ground truth text file should contain exactly one line, not %s" % (args.txt, len(lines)))
ValueError: ERROR: data/my-handwriting-ground-truth/1.gt.txt: Ground truth text file should contain exactly one line, not 1

stweil: > Was this value error supposed to be raised for lines greater than one?

Yes. The code here used a wrong check. That is fixed in the latest code.

PR #176, opened 2020-07-06
Always use bidi algorithm

stweil:
It is needed for RTL scripts and does not harm for LTR scripts.


stweil:
See also discussion for PR #159. This PR adds part of @Shreeshrii's PR without adding a new command line flag for RTL scripts.

PR #180, opened 2020-07-08
Integrate new bidi-library dependency

M3ssman:


.gitignore
```diff
@@ -18,3 +18,5 @@ data/*.traineddata
 wackenroder_herzensergiessungen_*.gt.txt
 wackenroder_herzensergiessungen_*.tif
 master.zip
+.vscode
```
stweil, line 21: Both additions are not necessary by the project, but only useful for some users. Such entries can be made in your local `.git/info/exclude` or in your global Git configuration. I prefer that. Otherwise .gitignore typically grows with more and more unrelated entries.

generate_wordstr_box.py
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 
 import argparse
-import bidi.algorithm
 import io
 import unicodedata
+
```
stweil, line 6: Some Python coding standards and their related formatters sort the import statements alphabetically. That's why I did likewise. What was the motivation for moving the import statement here to a different line?

stweil: Sorry, I should have looked at your commit message first. So pylint groups import statements, starting with standard imports, then third party imports. So that's fine. `isort` also implements that order.

stweil:
I'd merge this PR without the .gitignore commit.

.gitignore
M3ssman: I didn't realize this before. Thanks for the hint, I will follow your suggestion!

stweil:
Thank you, merged now. I removed the two commits for `.gitignore` from the git history.

PR #185, opened 2020-08-18
Bug Fix: generate_line_box.py should NOT throw an error when the number lines in GT file is equal to one

abhishekthanki:


kba:
Eek, good catch! @abhishekthanki I think the same problem exists in the other generate_*.py files. Can you fix the comparison there as well, please? Thanks.

abhishekthanki:
@kba Sure!

abhishekthanki:
@kba Done!

stweil (approved):
Thanks a lot.

PR #205, opened 2020-11-18
Feat/generate trainingsets

M3ssman:
Include generation of Trainingdata Sets from OCR like ALTO V3, PAGE 2013, PAGE 2019 and Image Files (tif, jpeg)


sets/training_sets.py
```diff
@@ -0,0 +1,336 @@
...
+        """Return bounding box of TexLine shape"""
+
+    def get_textline_content(self) -> str:
+        """
+        Set TextLine contents from it's included word tokens
+        may revert order of tokens if required
+        """
+
+        if self.revert:
+            return reduce(lambda c, p: p + ' ' + c, self.text_words)
```
kba, line 65: Since we already require python-bidi, it would probably more robust to use it for handling the inversion, c.f. https://github.com/MeirKriheli/python-bidi/blob/master/bidi/algorithm.py / https://github.com/MeirKriheli/python-bidi#api

kba (commented):
I haven't tested it yet, but looks very promising. I particularly appreciate the unit tests 👍

generate_sets.py
```diff
@@ -0,0 +1,78 @@
...
+APP_ARGUMENTS.add_argument(
+    "-o",
+    "--output",
+    required=False,
+    help="path output data")
+APP_ARGUMENTS.add_argument(
+    "-m",
+    "--minchars",
+    required=False,
+    help="Minimum chars for a line")
```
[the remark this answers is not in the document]
M3ssman: Historical note: Originates from newspaper digitalization project, where a common text line (no adds) has usually more than 20 chars. In fact, we only took care for lines with at least 32 chars, since I thought these lines were more valuable for training than shorter lines because of having more characters to learn.
But personally I'm totally free about this, so what about, say, 4 chars?

requirements.txt
```diff
@@ -1,2 +1,9 @@
 Pillow>=6.2.1
 python-bidi>=0.4
+lxml>=4.0.0
+exifread>=2.3.0
+opencv-python>=4.4.0
```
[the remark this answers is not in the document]
M3ssman: Argh, you're totally right! Since we do *not* want to display or plot something there's no need for the full stack.

sets/training_sets.py
M3ssman: Thanks for the hint, I'll take a look!

kba:
I've tested it now, unit tests pass and I managed to extract image-text pairs from the kant_aufklaerung_1784 sample in assets:

```sh
$ python3 ./generate_sets.py -d ../assets/data/kant_aufklaerung_1784/data/OCR-D-GT-PAGE/PAGE_0017_PAGE.xml -i ../assets/data/kant_aufklaerung_1784/data/OCR-D-IMG/INPUT_0017.tif 
[SUCCESS] created '20' training data sets, please review
```

It would be useful to make `-o` required or at least print the output directory as part of the `SUCCESS` message.

Could the `-i` argument be optional and by default be derived from `imageFilename` (PAGE) / `sourceImageInformation/filename` (ALTO)?

We also need a section on at least the CLI usage in the README.md

generate_sets.py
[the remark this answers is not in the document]
kba: It makes sense to skip short lines for training, but the fact that there is a minimum of chars and what that minimum is should be clearly communicated to the user, so he/she isn't surprised why some lines are skipped.

And yes, probably something low like 4 (documented in `./generate_sets.py --help` would be best IMHO.

sets/training_sets.py
M3ssman: Hm, 
I guess we need to go without bidi so far, since it looks the output from mixed arab + latin lines makes them turn from rtl to ltr. Lines with only arab chars and indic numbers seem to work pretty well with bidi, but mixed don't.

I'm no export on that, though. Please, take a look yourself. I've added the bidi import, adopted line content generation (the commented section). Feel free to switch implementations. (in your preferred IDE place a debugging mark in `test_create_sets_from_page2013_and_jpg` to inspect the temporary test files written)

M3ssman: Do I get it right: bidi works on char-Level? If so, I don't think it is useful in this scenario. I only know some (rather poor) arabic output generated from tesseract itself which is word-based.

M3ssman:
For the arabic text that is included as text resource (288652), and that's causing trouble with bidi, please see the original image (binarized)

generate_sets.py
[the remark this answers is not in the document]
stweil: It probably makes sense not to skip short lines for training. Tesseract was trained only with artifical long lines initially, and the standard models have problems with short lines (typically the page numbers, but also short lines ending paragraphs). We know that there exist valid lines with only a single character for page numbers (1 … 9, a … z, A … Z). Why should be skip lines with one or two characters as long as they are valid?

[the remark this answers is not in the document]
M3ssman: This originates from the decision to prefer longer lines because they provide more characters. I thought more characters means more training character training material, and more material increases pattern recognition accuracy. But this doesn't pay much attention to a characters' context. In newspapers advertises I've seen many lines that are way shorter than 8 chars, only containing abbreviations and alike. Maybe focusing on "regular article lines" is another reason why Tesseract (4.1.1) usually performs rather worse in this realm, compared to single-column-liners. 

@stweil Do you suggest to turn the minchars arg into being completely optional, or to set the default value to "1", to skip lines that only contain non-printable characters?

Shreeshrii:
@kba Do you know of any Devanagari or any other Indic language datasets in Page XML format? I only have scanned page images and and their groundtruth in text format. I don't think those will work with this PR.

kba:
> @kba Do you know of any Devanagari or any other Indic language datasets in Page XML format? I only have scanned page images and and their groundtruth in text format. I don't think those will work with this PR.

Sorry, I do not. But maybe you have OCR results in Devanagari to test the mechanics of this PR? What problems do you foresee with Devanagari?

Shreeshrii:
> What problems do you foresee with Devanagari?

I don't foresee any, but wanted to test with complex scripts, just in case there is any difference in processing.

> maybe you have OCR results in Devanagari to test the mechanics of this PR? 

Good idea. I can test using ALTO output from tesseract.

> Devanagari or any other Indic language datasets in Page XML format

I found a set of files at https://github.com/ramayanaocr/ocr-comparison/tree/master/Transkribus/Input, which has the png files as well as the xml files (generated by transkribus, I guess). I tested with one of those files, while the console messages reported success, the files were not created. The summary option created a file, but the file had a empty lines.

```
 tesstrain-extract-gt  /home/ubuntu/ocr-comparison/Transkribus/Input/page/ram110.xml -i /home/ubuntu/ocr-comparison/Transkribus/Input/ram110.png
[INFO   ] generate trainingsets of '/home/ubuntu/ocr-comparison/Transkribus/Input/page/ram110.xml' with '/home/ubuntu/ocr-comparison/Transkribus/Input/ram110.png' (min: 1, sum: False, reorder: False)
[SUCCESS] created '24' training data sets in 'training_data_ram110', please review
```

I tested with the Arabic image shared earlier in this thread with its xml file in resources, just to make sure that I had the PR installed correctly. That worked i.e. created the files. I haven't looked at the text within them.

```
tesstrain-extract-gt /home/ubuntu/tesstrain/tests/resources/xml/288652.xml -i /home/ubuntu/pagedeva/288652.png -o /home/ubuntu/pagedeva/output -s
[INFO   ] generate trainingsets of '/home/ubuntu/tesstrain/tests/resources/xml/288652.xml' with '/home/ubuntu/pagedeva/288652.png' (min: 1, sum: True, reorder: False)
[SUCCESS] created '33' training data sets in '/home/ubuntu/pagedeva/output', please review
```

Is there a compatibility issue with transkribus generated PAGE files?

Shreeshrii:
I tested just now with ALTO output from tesseract and get the following error:

```
tesstrain-extract-gt /home/ubuntu/tesstrain-San/test/iast/sandocs.xml -i /home/ubuntu/tesstrain-San/test/iast/sandocs.png -s
[INFO   ] generate trainingsets of '/home/ubuntu/tesstrain-San/test/iast/sandocs.xml' with '/home/ubuntu/tesstrain-San/test/iast/sandocs.png' (min: 1, sum: True, reorder: False)
/home/ubuntu/miniforge3/lib/python3.7/site-packages/numpy/core/_methods.py:234: RuntimeWarning: Degrees of freedom <= 0 for slice
  keepdims=keepdims)
/home/ubuntu/miniforge3/lib/python3.7/site-packages/numpy/core/_methods.py:195: RuntimeWarning: invalid value encountered in true_divide
  arrmean, rcount, out=arrmean, casting='unsafe', subok=False)
/home/ubuntu/miniforge3/lib/python3.7/site-packages/numpy/core/_methods.py:226: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
  File "/home/ubuntu/miniforge3/bin/tesstrain-extract-gt", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/cli.py", line 74, in main
    reorder=REORDER)
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/training_sets.py", line 358, in create
    prefix=prefix)
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/training_sets.py", line 268, in write_data
    image_handle, training_data)
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/training_sets.py", line 294, in _extract_frame
    return image_handle[start_vpos:end_vpos, start_hpos:end_hpos]
TypeError: 'NoneType' object is not subscriptable
```

M3ssman:
@Shreeshrii Thanks for pointing to PAGE-Files that miss `Word' elements at all! 
- Since that was the cause for the missing results in the provided Devanagari sample. I tried to fix this and integrated the file as new test resource. Unfortunately, I can't say a word about the textual outcome, so please update the PR and have a look again ...

Shreeshrii:
@M3ssman I tried just now but am getting the same result as before.

```
 git log -3
commit 3fb94996ac42818b302850080a6f2535db12251e (HEAD -> pagesets)
Author: M3ssman <uwe.hartwig@bitsrc.info>
Date:   Sun Dec 13 10:44:47 2020 +0100

    [app][fix] handle page without word elements

commit 2f3566bc23a848e3df7801b2fa1a6ce1d417e7bc
Author: M3ssman <uwe.hartwig@bitsrc.info>
Date:   Mon Dec 7 14:19:58 2020 +0100

    [app][fix] filter invalid lines

commit 57ba229ace0c9ae74afb889916cba3555ef7b4d0
Author: M3ssman <uwe.hartwig@bitsrc.info>
Date:   Mon Dec 7 13:18:48 2020 +0100

    [app][test] fix test imports
```

```
 tesstrain-extract-gt  /home/ubuntu/ocr-comparison/Transkribus/Input/page/ram110.xml -i /home/ubuntu/ocr-comparison/Transkribus/Input/ram110.png -s
[INFO   ] generate trainingsets of '/home/ubuntu/ocr-comparison/Transkribus/Input/page/ram110.xml' with '/home/ubuntu/ocr-comparison/Transkribus/Input/ram110.png' (min: 1, sum: True, reorder: False)
[SUCCESS] created '24' training data sets in 'training_data_ram110', please review
```

However, only the summary file in created in 'training_data_ram110'. File is attached.

[ram110_summary.gt.txt](https://github.com/tesseract-ocr/tesstrain/files/5689784/ram110_summary.gt.txt)

Shreeshrii:
I also tried with the ALTO 4.1 XML referenced in https://github.com/OCR-D/ocrd_fileformat/issues/23
That fails with the following messages:

```
(base) ubuntu@tesseract-ocr-1:~/tesstrain-pagesets$ tesstrain-extract-gt /home/ubuntu/OCR_GS_Data/TypeFaces/persian_watts_typeface/data/ahsan_at_tavarikh_31.xml -i /home/ubuntu/OCR_GS_Data/TypeFaces/persian_watts_typeface/data/ahsan_at_tavarikh_31.png -s
[INFO   ] generate trainingsets of '/home/ubuntu/OCR_GS_Data/TypeFaces/persian_watts_typeface/data/ahsan_at_tavarikh_31.xml' with '/home/ubuntu/OCR_GS_Data/TypeFaces/persian_watts_typeface/data/ahsan_at_tavarikh_31.png' (min: 1, sum: True, reorder: False)
  File "/home/ubuntu/miniforge3/bin/tesstrain-extract-gt", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/cli.py", line 74, in main
    reorder=REORDER)
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/training_sets.py", line 351, in create
    self.xml_data, min_len=min_chars, reorder=reorder)
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/training_sets.py", line 184, in text_line_factory
    ns_prefix = _determine_namespace(xml_data)
  File "/home/ubuntu/miniforge3/lib/python3.7/site-packages/generate_sets/training_sets.py", line 223, in _determine_namespace
    return [k for (k, v) in XML_NS.items() if v == root_tag][0]
IndexError: list index out of range
```

M3ssman:
@Shreeshrii Thanks for pointing towards ALTO V4. I've missed this before, since we're using the latest official stable release, tesseract 4.1., which doesn't create this kind of ALTO data. I've added the ALTO V4 namespace declaration and it worked fine. Somehow, I found this surprising, since the ALTO V4 data from OpenITI you pointed out looks quite unfamiliar, having String CONTENT spanned over a complete textline. I've never seen this before. Where does this data come from?

Regarding the Devanagari Issue: Your git log looks well, the version matches. Maybe `tesstrain-extract-gt` in your current, active environment is outdated, so please drop it and do a fresh install afterwards. You can also do a `pytest -v` to run the so far included test cases (with their test datasets) and check the temporary outputs in your local `/tmp/pytest-of-<account>` dir.

Shreeshrii:
```
pytest -v
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.7.6, pytest-6.2.0, py-1.10.0, pluggy-0.13.1 -- /home/ubuntu/miniforge3/bin/python3.7
cachedir: .pytest_cache
rootdir: /home/ubuntu/tesstrain-pagesets
collected 8 items

tests/test_generate_sets.py::test_create_sets_from_alto_and_tif PASSED                                                                                                                                                                [ 12%]
tests/test_generate_sets.py::test_create_sets_from_page2013_and_jpg PASSED                                                                                                                                                            [ 25%]
tests/test_generate_sets.py::test_create_sets_from_page2013_and_jpg_no_summary PASSED                                                                                                                                                 [ 37%]
tests/test_generate_sets.py::test_create_sets_from_page2019_and_png PASSED                                                                                                                                                            [ 50%]
tests/test_generate_sets.py::test_create_sets_from_ocrd_workdspace PASSED                                                                                                                                                             [ 62%]
tests/test_generate_sets.py::test_create_sets_from_ocrd_workdspace_fails PASSED                                                                                                                                                       [ 75%]
tests/test_generate_sets.py::test_handle_invalid_coords PASSED                                                                                                                                                                        [ 87%]
tests/test_generate_sets.py::test_handle_page_devanagari_with_texlines PASSED                                                                                                                                                         [100%]

============================================================================================================= warnings summary ==============================================================================================================
tests/test_generate_sets.py::test_create_sets_from_alto_and_tif
  /home/ubuntu/miniforge3/lib/python3.7/site-packages/numpy/core/_methods.py:234: RuntimeWarning: Degrees of freedom <= 0 for slice
    keepdims=keepdims)

tests/test_generate_sets.py::test_create_sets_from_alto_and_tif
  /home/ubuntu/miniforge3/lib/python3.7/site-packages/numpy/core/_methods.py:195: RuntimeWarning: invalid value encountered in true_divide
    arrmean, rcount, out=arrmean, casting='unsafe', subok=False)

tests/test_generate_sets.py::test_create_sets_from_alto_and_tif
  /home/ubuntu/miniforge3/lib/python3.7/site-packages/numpy/core/_methods.py:226: RuntimeWarning: invalid value encountered in double_scalars
    ret = ret.dtype.type(ret / rcount)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
====================================================================================================== 8 passed, 3 warnings in 12.88s =======================================================================================================

(base) ubuntu@tesseract-ocr-1:~/tesstrain-pagesets$ tesstrain-extract-gt  /home/ubuntu/ocr-comparison/Transkribus/Input/page/ram110.xml -i /home/ubuntu/ocr-comparison/Transkribus/Input/ram110.png -s
[INFO   ] generate trainingsets of '/home/ubuntu/ocr-comparison/Transkribus/Input/page/ram110.xml' with '/home/ubuntu/ocr-comparison/Transkribus/Input/ram110.png' (min: 1, sum: True, reorder: False)
[SUCCESS] created '24' training data sets in 'training_data_ram110', please review

(base) ubuntu@tesseract-ocr-1:~/tesstrain-pagesets$ ls -l training_data_ram110
total 4
-rw-rw-r-- 1 ubuntu ubuntu 24 Dec 15 04:20 ram110_summary.gt.txt
```

Shreeshrii:
The files are generated as part of the test:

```
(base) ubuntu@tesseract-ocr-1:/tmp/pytest-of-ubuntu/pytest-current/test_handle_page_devanagari_wicurrent$ ls -l
total 34492
-rw-rw-r-- 1 ubuntu ubuntu 22618835 Dec 15 04:18 ram110.png
-rw-rw-r-- 1 ubuntu ubuntu     2515 Dec 15 04:18 ram110_summary.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu      187 Dec 15 04:18 ram110_tl_10.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   603624 Dec 15 04:18 ram110_tl_10.tif
-rw-rw-r-- 1 ubuntu ubuntu       37 Dec 15 04:18 ram110_tl_11.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   266846 Dec 15 04:18 ram110_tl_11.tif
-rw-rw-r-- 1 ubuntu ubuntu      117 Dec 15 04:18 ram110_tl_12.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   550042 Dec 15 04:18 ram110_tl_12.tif
-rw-rw-r-- 1 ubuntu ubuntu      108 Dec 15 04:18 ram110_tl_13.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   601434 Dec 15 04:18 ram110_tl_13.tif
-rw-rw-r-- 1 ubuntu ubuntu      151 Dec 15 04:18 ram110_tl_14.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   651804 Dec 15 04:18 ram110_tl_14.tif
-rw-rw-r-- 1 ubuntu ubuntu      102 Dec 15 04:18 ram110_tl_15.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   520708 Dec 15 04:18 ram110_tl_15.tif
-rw-rw-r-- 1 ubuntu ubuntu      102 Dec 15 04:18 ram110_tl_16.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   516418 Dec 15 04:18 ram110_tl_16.tif
-rw-rw-r-- 1 ubuntu ubuntu      107 Dec 15 04:18 ram110_tl_17.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   745854 Dec 15 04:18 ram110_tl_17.tif
-rw-rw-r-- 1 ubuntu ubuntu      148 Dec 15 04:18 ram110_tl_18.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   615958 Dec 15 04:18 ram110_tl_18.tif
-rw-rw-r-- 1 ubuntu ubuntu      157 Dec 15 04:18 ram110_tl_19.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   560244 Dec 15 04:18 ram110_tl_19.tif
-rw-rw-r-- 1 ubuntu ubuntu       43 Dec 15 04:18 ram110_tl_1.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   127490 Dec 15 04:18 ram110_tl_1.tif
-rw-rw-r-- 1 ubuntu ubuntu      106 Dec 15 04:18 ram110_tl_20.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   561928 Dec 15 04:18 ram110_tl_20.tif
-rw-rw-r-- 1 ubuntu ubuntu      107 Dec 15 04:18 ram110_tl_21.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   662002 Dec 15 04:18 ram110_tl_21.tif
-rw-rw-r-- 1 ubuntu ubuntu      127 Dec 15 04:18 ram110_tl_22.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   548104 Dec 15 04:18 ram110_tl_22.tif
-rw-rw-r-- 1 ubuntu ubuntu      115 Dec 15 04:18 ram110_tl_23.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   704092 Dec 15 04:18 ram110_tl_23.tif
-rw-rw-r-- 1 ubuntu ubuntu       17 Dec 15 04:18 ram110_tl_24.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   105892 Dec 15 04:18 ram110_tl_24.tif
-rw-rw-r-- 1 ubuntu ubuntu        6 Dec 15 04:18 ram110_tl_2.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu    32458 Dec 15 04:18 ram110_tl_2.tif
-rw-rw-r-- 1 ubuntu ubuntu      137 Dec 15 04:18 ram110_tl_3.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   741314 Dec 15 04:18 ram110_tl_3.tif
-rw-rw-r-- 1 ubuntu ubuntu      145 Dec 15 04:18 ram110_tl_4.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   712610 Dec 15 04:18 ram110_tl_4.tif
-rw-rw-r-- 1 ubuntu ubuntu       36 Dec 15 04:18 ram110_tl_5.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   337642 Dec 15 04:18 ram110_tl_5.tif
-rw-rw-r-- 1 ubuntu ubuntu       99 Dec 15 04:18 ram110_tl_6.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   495246 Dec 15 04:18 ram110_tl_6.tif
-rw-rw-r-- 1 ubuntu ubuntu       97 Dec 15 04:18 ram110_tl_7.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   581738 Dec 15 04:18 ram110_tl_7.tif
-rw-rw-r-- 1 ubuntu ubuntu      103 Dec 15 04:18 ram110_tl_8.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   518032 Dec 15 04:18 ram110_tl_8.tif
-rw-rw-r-- 1 ubuntu ubuntu      137 Dec 15 04:18 ram110_tl_9.gt.txt
-rw-rw-r-- 1 ubuntu ubuntu   761348 Dec 15 04:18 ram110_tl_9.tif
-rw-rw-r-- 1 ubuntu ubuntu    22586 Dec 15 04:18 ram110.xml
```

How do I ensure that latest `tesstrain-extract-gt` is being used?

Shreeshrii:
The image should look like the following. But, in `/tmp/pytest-of-ubuntu/pytest-current/test_handle_page_devanagari_wicurrent` the png file as well as the generated tifs have ??? rather than the Devanagari text as per image.

ram110

**The generated gt.txt is correct (i.e. it is in Devanagari script) but the images are not.**

[ram110_summary.gt.txt](https://github.com/tesseract-ocr/tesstrain/files/5693404/ram110_summary.gt.txt)

Shreeshrii:
>ALTO V4 data from OpenITI you pointed out looks quite unfamiliar, having String CONTENT spanned over a complete textline. I've never seen this before. Where does this data come from?

I do not know more than the info available online. Please see
https://github.com/OpenITI/RELEASE
and
https://zenodo.org/record/4075046#.X9hC0dgzaUk

M3ssman:
@Shreeshrii Please note, test images are just created on-the-fly, with a library that is out-of-the-box just able to render a very small subset of UTF-8 chars, I guess only ASCII, neither arabic, persian,  devanagari or old german fracture letters. This was introduced to keep test data small and free from binary image stuff. It only gives you a hint whether the lines would match the "words".

M3ssman:
@Shreeshrii Regarding the lastest version: currently, there's only a-pre-beta-version (0.0.1) annotated in the `setup.py`. Usually this would be the place to follow versioning. I do not know how to utilize some sort of repository information straight at this point. Maybe @kba can give us a hint?

Shreeshrii:
@M3ssman Thanks for the explanations regarding test files. 

I removed `tesstrain-extract-gt` from the `bin` directories and reinstalled in the environment where `ocrd` is installed. It works now. All the tif and gt.txt were created for the Transkribus Devanagari file. 

The alto4.1 Persian file is also working. This is great!! Thank you.

M3ssman:
@Shreeshrii You're welcome!

... Sorry for the confusion regarding RTL ... finally, it turned out that the `-r` flag aims at something different than _real_ RTL which can be handled with `py-bidi`. If active, it only re-arranges word tokens by top-left-corner in descending order, starting from right margin. Therefore I renamed it to `--reorder`. It *doesn't* turn characters. I had to deal with arabic PAGE-XML exported from Transkribus, having inconsistent reading-orders and display artifacts and almost made me go crazy. 

Since this relies on individual coordinates for each token, I'm afraid it will have no effect on test resources like the ones gathered from OpenITI which only have a single `String@CONTENT` element that represents a text line in total (or at least more than just one word). Reordering this way requires proper coordinates below text line level: We can't just chop the lines and reorder tokens, since the source order of elements of a plain text line is certainly not always reliable.

Shreeshrii:
This should not be closed. It needs review by someone familiar with RTL languages.

generate_sets.py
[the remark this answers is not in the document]
stweil: Setting minchars to 1 sounds reasonable. I cannot imagine how a line which only contains non-printable characters would look like.

M3ssman:
I've been talking with https://github.com/galdring , a colleague, about this review and he's out to get us somebody.

tests/test_generate_sets.py
```diff
@@ -0,0 +1,403 @@
...
+@pytest.fixture(name='fixture_page2013_jpg')
+def _fixture_page2013_jpg(tmpdir):
+
+    res = os.path.join(RES_ROOT, 'xml', OCR_TRANSK_DATA)
+    path_page = tmpdir.mkdir('training').join(OCR_TRANSK_DATA)
+    shutil.copyfile(res, path_page)
+
+    words = extract_words(path_page)
+
+    file_path = tmpdir.mkdir('images').join('288652.jpg')
```
stweil, line 149:
```suggestion
    file_path = tmpdir.mkdir('images').join(OCR_TRANSK_IMAG)
```

tests/test_generate_sets.py
```diff
@@ -0,0 +1,403 @@
...
+        folder_out=path_input_dir)
+
+    # assert
+    expected_len = 33
+    assert len(data) == expected_len
+    path_items = os.listdir(os.path.dirname(fixture_page2019_png))
+    tifs = [tif for tif in path_items if str(tif).endswith(".tif")]
+    assert len(tifs) == expected_len
+
+    txt_files = [txt for txt in path_items if str(txt).endswith(".gt.txt")]
```
stweil, line 254:
```suggestion
    txt_files = [txt for txt in path_items if str(txt).endswith(GT_SUFFIX)]
```

stweil:
@M3ssman, please check `git config user.name`. Your commits use that name for the author information.

M3ssman:
There's also a branch with the same name (`feat/generate-trainingsets`) but outdated already in this repository, which I guess @kba created to commit his extensions before I integrated them and they finally went to `ulb-sachsen-anhalt/tesstrain/tree/feat/generate-trainingsets`. 
I wonder if this causes any irritations?

kba:
> I wonder if this causes any irritations?

I don't think so but I deleted the branch since it is outdated as you say.

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+import os
+from pathlib import (
+    Path
+)
+from functools import reduce
+import sys
+
+import exifread
+import lxml.etree as etree
+import numpy as np
```
stweil, line 14:
```suggestion
```

stweil: This fixes a warning from LGTM.

generate_sets.py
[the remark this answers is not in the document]
bertsky: I agree with Stefan we should make minchars optional and try to make Tesseract learn short lines well. Not sure how the LSTM implementation here unrolls, but short lines should create fewer weight updates, so characters would still contribute "democratically" – just that there's more incentive to get a better transition from the initial state.

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+            for mark in CLEAR_MARKS:
+                if mark in strip:
+                    self.text_words[i] = strip.replace(mark, '')
+
+    @staticmethod
+    def _pick_top_left(elem, namespace):
+        coords = elem.find(f'{namespace}:Coords', XML_NS)
+        points = coords.attrib['points'].split()
+        if len(points) > 0:
+            return points[0].split(',')[0]
```
bertsky, line 183: That's not the top left – that's the first point pair's x coordinate (which could be anywhere in the polygon). 

It's probably not an issue if Words do not overlap, but why not just use the center, just to be on the safe side? 

Also, what about vertical text here? In that case you need to sort by y, not by x.

```suggestion
    def _pos(elem, namespace, vertical=False):
        coords = elem.find(f'{namespace}:Coords', XML_NS)
        points = coords.attrib['points'].split()
        center = tuple(map(lambda c: sum(c) / len(c), zip(*points)))
        if vertical:
            return center[1]
        return center[0]
```

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+        self.set_text()
+        if self.valid:
+            self.shape = self.get_shape(self.element)
+
+    def set_id(self):
+        self.element_id = self.element.attrib['ID']
+
+    def set_text(self):
+        strings = self.element.findall(f'{self.namespace}:String', XML_NS)
+        self.text_words = [e.attrib['CONTENT'] for e in strings]
```
bertsky, line 105: What about the final hyphenation character if it exists?

```suggestion
        strings = self.element.findall(f'{self.namespace}:String', XML_NS)
        strings.extend(self.element.findall(f'{self.namespace}:HYP', XML_NS)
        self.text_words = [e.attrib['CONTENT'] for e in strings]
```

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+
+        p_attr = element.find(
+            f'{self.namespace}:Coords',
+            XML_NS).attrib['points']
+        numbers = [int(n) for pair in p_attr.split() for n in pair.split(',')]
+
+        # group clustering idiom
+        points = list(zip(*[iter(numbers)] * 2))
+
+        return Polygon(points)
```
bertsky, line 199: That does not make sense to me. Why do you ungroup the points, only to regroup them right away?

```suggestion
        return Polygon(tuple(map(int, pair.split(','))) for pair in p_attr.split())
```

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+            if text_equiv.text:
+                stripped = text_equiv.text.strip()
+                if len(stripped) and len(stripped) >= min_len:
+                    matchings.append(textline)
+            else:
+                words = textline.findall(
+                    f'{ns_prefix}:Word/{ns_prefix}:TextEquiv/{ns_prefix}:Unicode', XML_NS)
+                if len(words):
+                    msg = f"[{xml_data.base}] no text but words for line '{textline.attrib['id']}'"
+                    raise RuntimeError(msg)
```
bertsky, line 227: Isn't this overly restrictive? You are preferring Word content in PageLine anyway, so why require TextEquiv on the TextLine level? You could just verify all the Word contents sum up to min_len here, like above.

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+
+    @staticmethod
+    def _extract_frame(image_handle, text_line):
+        the_shape = text_line.shape
+        the_box = the_shape.bounds
+        start_vpos = int(the_box[1])
+        end_vpos = int(the_box[3])
+        start_hpos = int(the_box[0])
+        end_hpos = int(the_box[2])
+        return image_handle[start_vpos:end_vpos, start_hpos:end_hpos]
```
bertsky, line 329: This unnecessarily throws away any polygon segmentation and just uses the bounding boxes for cropping. This could be problematic if there are overlaps between bounding boxes (as in densely typeset historic texts, or coarse line segmentation). Of course, tesstrain could still be run with PSM=7 or such, but why not use the polygon for masking? ALTO v4 has `Shape/Polygon/POINTS`, and PAGE of course is always polygonal.

The only problem with masking could be the background colour: If the image is not binarized when imreading, simly using white could create an artificial "edge". But you can easily use PIL.ImageStat.Stat's or Numpy's `median` here. See `ocrd_utils.image_from_polygon` for a full solution.

bertsky: Also, it could be useful to look for an `AlternativeImage` on the line level instead of cropping here.

extract_sets/training_sets.py
```diff
@@ -0,0 +1,398 @@
...
+                    cv2.IMWRITE_TIFF_YDPI, self.ydpi]
+        return []
+
+    def _read_dpi(self):
+        if str(self.path_image_data).endswith(".tif"):
+            (self.xdpi, self.ydpi) = TrainingSets.read_dpi_from_tif(
+                self.path_image_data)
+        elif str(self.path_image_data).endswith(".jpg"):
+            (self.xdpi, self.ydpi) = TrainingSets.read_dpi_from_jpg(
+                self.path_image_data)
```
bertsky, line 348: How about DPI metadata in PNGs?

extract_sets/training_sets.py
M3ssman: Sorry for the irritation, this was rather a WIP commit

extract_sets/training_sets.py
M3ssman: Sounds reasonable, I will do some tests this way

extract_sets/training_sets.py
M3ssman: You're right, I didn't take this into account yet.

extract_sets/training_sets.py
M3ssman: Actually, I did it this way to make clear that there are two distinct actions to take: first parse the textual information from the coords attribute, second generate a list of int pairs that finally form the points. To separate them doesn't look as elegant as your suggestion, but it's easier to understand.

extract_sets/training_sets.py
M3ssman: I've seen some samples of broken data, therefore the paranoia

extract_sets/training_sets.py
M3ssman: Good point! I've just missed this ...

extract_sets/training_sets.py
M3ssman: You're right, I'm still not finished to work this out.

extract_sets/training_sets.py
bertsky: I see. Point taken that the one-liner is not as accessible, but ungrouping is not necessary, so that's not very understandable either IMHO. I was led to believe there's more going on initially, because of the docstring says
> Coordinate data from current OCR-D-Workflows can contain lots of points, therefore additional calculations are required

So this was probably more a TODO than a description of the current implementation? (You could use Shapely's `Polygon.simplify(tolerance)` if you want fewer points, or do lots of other stuff to get a valid non-intersecting linear ring...)

extract_sets/training_sets.py
bertsky: Broken in what way? PAGE-XML does permit a TextLine without TextEquiv having Words with TextEquiv. In the extraction, you even distrust the line-level text and concatenate from the word-level elements sorted by coordinates (which I agree is actually the right thing to do for OCR training, and much more robust than trusting `@readingDirection` and element ordering, especially with RTL and LTR/RTL mix). I guess it depends on the tool that produced the XML what you get.

extract_sets/training_sets.py
bertsky: > fd09bb0

Wow, that was super-fast! Will take me some time to re-review. Have never seen use of `cv2.fillConvexPoly` before, will try it out. Creating background with (blurred) random noise is a very interesting idea (but I'm afraid this can currently overflow and wrap-around).

extract_sets/training_sets.py
M3ssman: Yep, that's exactly the point. We are confronted (rather often) with so-called PAGE-XML from external sources produced by unknown Tools. It's rumored to be done with some Transkribus of yore, but it looks like a human fumbled straight into XML.

bertsky: But that's exactly why I am wondering here what makes you give up with an exception. Why write that robust Word/TextEquiv extractor if you are not using it when it's needed the most?

M3ssman: The Problem is a *real* Problem just in rather exotic circumstances. We were confronted with Transkriptions in PAGE exported from Transkribus with only lines, no words, therefore missing proper word-coordinates for re-ordering.

extract_sets/training_sets.py
M3ssman: What about calculating a bounding rect and then picking the polygon point with shortest distance? 
In case of a *real* rectangular shape, this should be the top left corner. And no need for vertical flag.

bertsky: > What about calculating a bounding rect and then picking the polygon point with shortest distance?
> In case of a _real_ rectangular shape, this should be the top left corner. 

Much too complicated. And Shapely's distance/project is not as easy to use as it looks on first sight (trust me). Above code does the job nicely. Plus you get the option for vertical sorting for free.

> And no need for vertical flag.

With vertical text you need to sort in y direction, not in x direction.

zdenop:
@M3ssman: can you please update your PR to current git code (python code is in [src](https://github.com/tesseract-ocr/tesstrain/tree/main/src) see [Migrate Python code to a dedicated package](https://github.com/tesseract-ocr/tesstrain/pull/309))

M3ssman:
@zdenop Sorry for the late reply.

What layout do you prefer? 
`<project_root>/src/extract_sets`  or integrate `training_sets.py` somehow into `<project_root>/src`  as part of `<project_root>/src/tesstrain` ?

stefan6419846:
If I understood @zdenop correctly, the final goal is to make everything available through the *tesstrain* Python package in the end. As you provide a dedicated entry point, `src/tesstrain` sounds like the appropriate package.

Nevertheless, I am not sure about the external dependencies. They might should be made optional (`extras_require`).

M3ssman:
@stefan6419846 
Thanks for your reply! Do you suggest to push these dependencies into  setuptool.setup.extras_require?

stefan6419846:
@M3ssman If you are going to integrate the training set generator into the existing Python package, I would suggest yes. At least for me they appear to be overkill for most users which just want to use the basic artificial training functionality.

extract_sets/training_sets.py
```diff
@@ -0,0 +1,655 @@
...
+
+
+def clear_vertical_borders(image_frame, intrusion_ratio):
+    """
+    Clear vertical overlappings by:
+    * binarize blurred input image data
+    * drawing artificial border
+    * collect only contours that touch this
+    * get contours that are specific ratio to close to the edge
+    * fill those with specific grey tone
```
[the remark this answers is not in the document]
bertsky: Also, didn't you write a textured fill (`grey_canvas` IIRC) for that very purpose (but for synthetic training) already?

bertsky:
> Nevertheless, I am not sure about the external dependencies. They might should be made optional (`extras_require`).
> 
> At least for me they appear to be overkill for most users which just want to use the basic artificial training functionality.

I disagree with that assessment. The pkg for _synthetic training_ is as relevant as some way to import from the widely used file formats (ALTO, PAGE) for _real GT training_ IMO. So if the trainingsets extension is adopted (at all), then its dependencies should _not_ be moved to `extras_require`.

Issue #172, opened 2020-06-30
Can't encode transcription: adding 0 automatically at start of ground truth file.


jayawantkarale:
Can't encode transcription: ' 0अंशक्लृप्ति ( aṁśa-kḷpti) f. A division into parts औपाधिक्यंशक्लृप्तिर्घटग-' in language ''

Above error display during training process. while providing ground truth text we given input as 
'अंशक्लृप्ति ( aṁśa-kḷpti) f. A division into parts औपाधिक्यंशक्लृप्तिर्घटग-'  but during training process it append 0 at start of line.
so this also add o after performing ocr extraction using  newly genrated traineddata.
why it is adding 0 in start of ground truth line.
Please help me to resolve the issue.

wrznr:
@jayawantkarale Without sample data, it is not possible to assist you since we cannot reproduce the erroneous behavior. If you could provide us with a minimal example which causes your issue?

jayawantkarale:
@wrznr thanks for your prompt reply. This erroneous behavior appear while training on large dataset. it is not possible to upload whole data, I am trying to reproduce same behavior  on small sample data. so after getting error on small dataset i will upload sample data.

Shreeshrii:
Can't encode transcription: ' 0अंशक्लृप्ति ( aṁśa-kḷpti) f. A division into parts औपाधिक्यंशक्लृप्तिर्घटग-' in language ''

Try to find the above text in your large dataset. Then using uniview or similar utility check whether there is any unprintable/invisible character in the line.

jayawantkarale:
I am able to reproduce same behaviour on smaller dataset, on attaching dataset along with its log file.
Now it is attaching 0 at start of most of the lines, and gives error can't encode transcription.

Can't encode transcription: ' 0MahāBhā. xii. 47. 42; 6 A ii 1/124 of a day द्यु हेयं पर्व चेत्  पादे पादर्स्रिशत्तु' in language ''
page no : vol1_101_1-029.exp0.tif

Can't encode transcription: ' 0MahāBhā. ii. 8. 14; विजिती वीतिहोत्रोंऽशः [ v. 1 ºञश्च ] MahāBhā. i. 1. 173.' in language ''
page no : vol1_101_1-048.exp0.tif

Can't encode transcription: ' 0day as a meaning found in L. 9 adv. (aṁśena aṁśe) in part, partly,' in language ''
page no : vol1_101_2-004.exp0.tif

[...]

PR #215, opened 2020-12-16
Concatenate with newline for all-gt , remove sort uniq

Shreeshrii:
See https://github.com/tesseract-ocr/tesstrain/issues/172#issuecomment-655246018

For large amount of training data, the current implementation hangs at times.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ training: $(OUTPUT_DIR).traineddata
 
 $(ALL_GT): $(shell find $(GROUND_TRUTH_DIR) -name '*.gt.txt')
 	@mkdir -p $(OUTPUT_DIR)
-	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs cat | sort | uniq > "$@"
+	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs -I{} sh -c "cat {}; echo ''" > "$@"
 
 .PRECIOUS: %.box
 %.box: %.png %.gt.txt
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ training: $(OUTPUT_DIR).traineddata
 
 $(ALL_GT): $(shell find $(GROUND_TRUTH_DIR) -name '*.gt.txt')
 	@mkdir -p $(OUTPUT_DIR)
-	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs awk 'FNR==1{print ""}1' > "$@"
+	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs paste -s > "$@"
 
 .PRECIOUS: %.box
 %.box: %.png %.gt.txt
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ training: $(OUTPUT_DIR).traineddata
 
 $(ALL_GT): $(shell find $(GROUND_TRUTH_DIR) -name '*.gt.txt')
 	@mkdir -p $(OUTPUT_DIR)
-	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs -I{} sh -c "cat {}; echo ''" > "$@"
+	find $(GROUND_TRUTH_DIR) -name '*.gt.txt' | xargs awk 'FNR==1{print ""}1' > "$@"
 
 .PRECIOUS: %.box
 %.box: %.png %.gt.txt


kba (approved):

Shreeshrii:
Please merge this.

stweil (changes requested):
I did a test on a large data set (375771 lines):

    $ time (find data/gt -name "*.gt.txt"|xargs cat | sort | uniq >/tmp/old)
    real	0m4.944s
    user	0m3.626s
    sys	0m4.974s

    $ time (find data/gt -name "*.gt.txt"|xargs -I{} sh -c "cat {}; echo ''" >/tmp/new)
    # I stopped that after several minutes.

So the new code is more than a factor 100 slower than the old one. This is caused by the huge number of processes started with the new code.

Maybe we can write a specialized Python script which does the work faster.

Shreeshrii:
@stweil Thanks for pointing out the performance issue.

The problem I was trying to solve is related to having ground-truth text files without \n at end. `cat` was making a really huge single line with it and sort|uniq would hang.

Would using `paste` work?

(base) ubuntu@tesseract-ocr-1:~/tesstrain-San$ time (find data/San-ground-truth/test -name "*.gt.txt"|xargs cat | sort | uniq >/tmp/old)

real    0m0.780s
user    0m0.688s
sys     0m0.261s
(base) ubuntu@tesseract-ocr-1:~/tesstrain-San$ time (find data/San-ground-truth/test -name "*.gt.txt"|xargs -I{} sh -c "cat {}; echo ''" >/tmp/new)

real    0m15.562s
user    0m11.121s
sys     0m5.300s
(base) ubuntu@tesseract-ocr-1:~/tesstrain-San$ time (find data/San-ground-truth/test -name "*.gt.txt"| xargs paste -s -d \n > /tmp/paste)

real    0m0.224s
user    0m0.071s
sys     0m0.189s
(base) ubuntu@tesseract-ocr-1:~/tesstrain-San$ find data/San-ground-truth/test -name "*.gt.txt"|wc -l
5982

Shreeshrii:
`awk` maybe a better option.

```

find data/San-ground-truth/train -name "*.gt.txt"|wc -l
177605

time (find data/San-ground-truth/train -name "*.gt.txt"| xargs awk 'FNR==1{print ""}1' > /tmp/awk)

real    0m6.053s
user    0m2.374s
sys     0m6.745s

time (find data/San-ground-truth/train -name "*.gt.txt"|xargs cat | sort | uniq >/tmp/old)

real    0m20.928s
user    0m19.759s
sys     0m7.948s
```

stweil:
Would a simple `find data/San-ground-truth/train -name "*.gt.txt"|xargs cat >all-gt` work for you, too? Then I'd prefer that as an intermediate solution.

It should not be necessary to avoid long lines, as they don't matter for `unicharset_extractor`. I had added `sort` and `uniq` to reduce the size of the resulting `all-gt` file.

My ideal solution would be a Python script which processes the ground truth text lines, extracts all characters and either passes a sorted list of characters to `unicharset_extractor` or directly creates the `unicharset` file. That script could optionally also count the frequency of the different characters.

stweil:
`find data/San-ground-truth/train -name "*.gt.txt"|xargs paste -s >all-gt` works also good for me. It creates separate lines without adding unneeded line feeds, so the resulting `all-gt` is smaller than with `awk`.

Shreeshrii:
@stweil OP's report is https://github.com/tesseract-ocr/tesstrain/issues/172

[zip file with gt](https://github.com/tesseract-ocr/tesstrain/files/4883022/san9_test-ground-truth.zip) for testing

Log file showed:

```
Wrote unicharset file data/san9_test/my.unicharset
merge_unicharsets data/san/san9_test.lstm-unicharset data/san9_test/my.unicharset  "data/san9_test/unicharset"
Loaded unicharset of size 225 from file data/san/san9_test.lstm-unicharset
Loaded unicharset of size 3 from file data/san9_test/my.unicharset
Wrote unicharset file data/san9_test/unicharset.
unicharset of size 3 from file data/san9_test/my.unicharset
```

**unicharset of size 3 from file data/san9_test/my.unicharset**

feff (BOM) was part of many of the gt lines. 

When single line ground truth text is being concatenated, it becomes one huge line and so if there is any error in even one file, the unicharset generation fails. 

His dataset has gt.txt files with a single line, no linefeed. When transcription has newline at end, then the problem does not arise.

So, I would prefer a solution that adds the newlines, because I think that solves the problem. But there might be a better way.

Please test with his dataset and recommend the way forward.

stweil:
My favourite solution is `find data/San-ground-truth/train -name "*.gt.txt"|xargs paste -s >all-gt`, see my comment above.

Issue #157, opened 2020-04-16
High error rate on training with Impact on RTL language (Kur) Persian-Arabic script #151


sam-kurdi:
I am getting a very high error (85) rate after training with Impact
I have started the training by the following configuration :

tesseract version:

tesseract 4.0.0-beta.1
leptonica-1.75.3
libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.2) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
Found AVX2
Found AVX
Found SSE

1:the workflow is utilized https://github.com/tesseract-ocr/tesstrain  .  Command-line to run make file is (make training MODEL_NAME=krd START_MODEL=ara LANG_TYPE=RTL FINETUNETYPE=Impact)
2:I added inherited.unicharset , ara.config , kur langdata,  Arabic.unicharset, and Latin.unercharset  provided by https://github.com/tesseract-ocr/langdata_lstm
3:I used ara.traineddata as a start model from https://github.com/tesseract-ocr/tessdata_best
4: (1304 / 2) imagelines + ground truth transcription 

could you please tell is there any misconfiguration?
how can I improve the accuracy rate?

Training Log :

pc1@pc:~/Desktop/tesstrain-master$ make training MODEL_NAME=krd START_MODEL=ara LANG_TYPE=RTL FINETUNETYPE=Impact

find data/krd-ground-truth -name '*.gt.txt' | xargs cat | sort | uniq > "data/krd/all-gt"
combine_tessdata -u /home/pc1/Desktop/tesstrain-master/usr/share/tessdata/ara.traineddata  data/ara/krd
Extracting tessdata components from /home/pc1/Desktop/tesstrain-master/usr/share/tessdata/ara.traineddata
Wrote data/ara/krd.config
Wrote data/ara/krd.lstm
Wrote data/ara/krd.lstm-punc-dawg
Wrote data/ara/krd.lstm-word-dawg
Wrote data/ara/krd.lstm-number-dawg
Wrote data/ara/krd.lstm-unicharset
Wrote data/ara/krd.lstm-recoder
Wrote data/ara/krd.version
Version string:4.00.00alpha:ara:synth20170629:[1,48,0,1Ct3,3,16Mp3,3Lfys64Lfx96Lrx96Lfx512O1c1]
0:config:size=545, offset=192
17:lstm:size=11582395, offset=737
[...]

PR #223, opened 2020-12-25
From stweil's custom makefile - calculate MAX_ITERATIONS from EPOCHS

Shreeshrii:
See https://github.com/tesseract-ocr/tesstrain/issues/157#issuecomment-615873731 for the patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,8 +52,16 @@ TESSDATA_REPO = _best
 # Ground truth directory. Default: $(GROUND_TRUTH_DIR)
 GROUND_TRUTH_DIR := $(OUTPUT_DIR)-ground-truth
 
+# Epochs. Default: $(EPOCHS)
+EPOCHS :=
+
 # Max iterations. Default: $(MAX_ITERATIONS)
-MAX_ITERATIONS := 10000
+ifeq ($(EPOCHS),)
+ MAX_ITERATIONS := 10000
+else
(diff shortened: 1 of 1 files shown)


kba (approved):

stweil:
I did not suggest that modification myself because it requires an existing `list.train`. That means it is not a general solution, and it is very likely that people will get unexpected results.

So we need a better solution to really support epochs. Maybe all programs which use the iteration parameter (`lstmtraining`, others?) can be enhanced to allow either a number (for iterations) or a number followed by `epochs` (for epochs).

Shreeshrii:
Even if not merged, keeping this code in the PR makes it easier to find :-)

zdenop:
Please use wiki for taking notes not issue tracker or PR.

stweil:
Pull request https://github.com/tesseract-ocr/tesseract/pull/3236 allows a more stable implementation. With that change, we either need only an update of the `tesstrain` documentation because a negative value for `MAX_ITERATIONS` is then interpreted as epochs. Or we can add an `EPOCH` parameter which is forwarded to a negative `MAX_ITERATIONS`.

Shreeshrii:
>we can add an EPOCH parameter which is forwarded to a negative MAX_ITERATIONS.

I would prefer this option. Thanks!

Shreeshrii:
Closing, in favor of https://github.com/tesseract-ocr/tesstrain/pull/228

Issue #221, opened 2020-12-20
Feature Request: Character Frequency in Training Text


Shreeshrii:
@stweil How to get a report like `Analyse-Report Version 0.1` shown in [NZZ wiki page](https://github.com/tesseract-ocr/tesstrain/wiki/NZZ)?

kba:
Looks a bit like output of https://github.com/eddieantonio/ocreval or https://github.com/impactcentre/ocrevalUAtion but `PosixPath` implies it's a Python tool producing this.

stweil:
That report was generated by @JKamlah.

Shreeshrii:
Thanks. I have seen such reports as part of `accuracy` output from `ocreval`. Kraken generates them for both the training and testing sets. I think it will be useful to add it as part of tesstrain.

I found a [python script](https://github.com/cmroughan/kraken_generated-data/blob/master/tools/count_chars.py) which generates similar info. It is from `https://github.com/cmroughan/kraken_generated-data` in the tools directory.  https://github.com/wincentbalin/pytesstrain also has [some useful tools](https://github.com/wincentbalin/pytesstrain/tree/master/pytesstrain/cli) which generate a wordlist as well as unigram and bigram frequencies.

PR #235, opened 2021-03-16
Create Character Count from training text

Shreeshrii:
Character frequency report 
Source: https://github.com/cmroughan/kraken_generated-data/blob/master/tools/count_chars.py

USAGE: count_chars.py <txt_file> | sort -n -r > <txt_file>.charcount

### DIFF
--- a/count_chars.py
+++ b/count_chars.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# Source: https://github.com/cmroughan/kraken_generated-data/blob/master/tools/count_chars.py
+
+import unicodedata
+import sys, getopt
+
+def main(argv):
+    txt_file = ''
+    try:
+        opts, args = getopt.getopt(argv,"h")
+    except getopt.GetoptError:
+        print('USAGE: count_chars.py <txt_file>')
+        sys.exit(2)
+    for opt, arg in opts:
+        if opt in ('-h','--help'):
+            print('USAGE: count_chars.py <txt_file> | sort -n -r > <txt_file>.charcount')
+            sys.exit()
+    for arg in args:
+        txt_file = arg
+    
+    inFile = open(txt_file)
+    rawText = inFile.read()
+    inFile.close()
+
+    chars = {}
+    for char in rawText:
+        if char not in chars:
+            chars[char] = 1
+        else:
+            chars[char] +=1
+
+    keys = list(chars.keys())
+    keys.sort()
+    for char in keys:
+        try:
+            print(chars[char],'\t',char,'\t',
+                  unicodedata.name(char))
+        except:
+            pass
+
+if __name__ == "__main__":
+   main(sys.argv[1:])
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,9 @@ wackenroder_herzensergiessungen_*.tif
 master.zip
 plot/*.LOG
 plot/ocrd*
+data
+src/training/__pycache__
+plot/*.tsv
+plot/*.log
+plot/*.list
+src/training/tesstrain_utils-*.py


bertsky:
Fixes #221 IINM

count_chars.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# Source: https://github.com/cmroughan/kraken_generated-data/blob/master/tools/count_chars.py
```
kba, line 4: Note that this is GPL code. Not sure about the consequences.

kba (approved):
A useful script. Maybe add a sentence about it to the README so people know its there and/or a Makefile target for this?

bertsky (changes requested):
Useful indeed, but I wonder why they sort outside on the shell if we already run Python. 

Also, it would help to offer a rule for the makefile already.

Besides, in https://github.com/tesseract-ocr/tesstrain/pull/260/commits/314e799e8d5457d9792ee7e6b62b60b93036e64f I proposed a similar functionality (only using shell means, i.e. `grep -o . | sort | uniq -c | sort -rn`) – it does not show codepoint names via `unicodedata.name`, but apart from that should be the same).

Shreeshrii:
@bertsky I copied the script from another repo and added an external sort as a quick hack to get the char count. Please feel free to modify as needed.

count_chars.py
```diff
@@ -0,0 +1,44 @@
...
+
+import unicodedata
+import sys, getopt
+
+def main(argv):
+    txt_file = ''
+    try:
+        opts, args = getopt.getopt(argv,"h")
+    except getopt.GetoptError:
+        print('USAGE: count_chars.py <txt_file>')
```
bertsky, line 14:
```suggestion
        print(usage)
```

count_chars.py
```diff
@@ -0,0 +1,44 @@
...
+        sys.exit(2)
+    for opt, arg in opts:
+        if opt in ('-h','--help'):
+            print('USAGE: count_chars.py <txt_file> | sort -n -r > <txt_file>.charcount')
```
bertsky, line 18:
```suggestion
            print(usage)
```

count_chars.py
```diff
@@ -0,0 +1,44 @@
...
+            sys.exit()
+    for arg in args:
+        txt_file = arg
+    
+    inFile = open(txt_file)
+    rawText = inFile.read()
+    inFile.close()
```
bertsky, line 25:
```suggestion
    rawText = ''
    for arg in args:
        with open(arg) as fd:
            rawText.append(fd.read())
    if not args:
        rawText.append(sys.stdin.read())
```

count_chars.py
```diff
@@ -0,0 +1,44 @@
...
+
+    chars = {}
+    for char in rawText:
+        if char not in chars:
+            chars[char] = 1
+        else:
+            chars[char] +=1
+
+    keys = list(chars.keys())
+    keys.sort()
```
bertsky, line 35:
```suggestion
    if '-n' in opts or '--numeric' in opts:
        keys = sorted(chars.keys(), key=chars.get)
    else:
        # apply system locale
        locale.setlocale(locale.LC_ALL, '')
        keys = sorted(chars.keys(), key=locale.strxfrm)
```

count_chars.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Source: https://github.com/cmroughan/kraken_generated-data/blob/master/tools/count_chars.py
+
+import unicodedata
+import sys, getopt
+
+def main(argv):
+    txt_file = ''
+    try:
+        opts, args = getopt.getopt(argv,"h")
```
bertsky, line 12:
```suggestion
        opts, args = getopt.getopt(argv, "hn", ["help", "numeric"])
```

bertsky (commented):
This adds sorting (by current locale, or optionally by frequency), as well as multi-file input.

Still missing: makefile rule, using `$(ALL_GT)`. (I cannot use Github suggestions for this.)

count_chars.py
stweil: Using GPL code requires changing the license for tesstrain to GPL. I don't think that we should do this, so GPL code cannot be used here.

stweil (changes requested):
We cannot use GPL code in tesstrain with Apache license.

zdenop:
Please make a description of input data and needed output, so those of us how did not have a look at the file can create a new script under Apache license.

bertsky:
> We cannot use GPL code in tesstrain with Apache license.

I don't think this short script meets the threshold of originality, though. After all, it just counts characters of files in Python. My above suggestions already changed most of the script's lines to make it more useful and makefile-workable. Just apply them and strip the Kraken reference. (I am not contesting Kraken's originality, only that single file's.)

bertsky:
> Also, it would help to offer a rule for the makefile already.
> 
> Besides, in [314e799](https://github.com/tesseract-ocr/tesstrain/commit/314e799e8d5457d9792ee7e6b62b60b93036e64f) I proposed a similar functionality (only using shell means, i.e. `grep -o . | sort | uniq -c | sort -rn`) – it does not show codepoint names via `unicodedata.name`, but apart from that should be the same).

Or abandon this PR and just merge #260.

zdenop:
Here is my python solution without need of extra tools. I did not implement reading form `stdin` as I do not see it usage in "make training"....

Usage: `python3 count_chars.py data/foo/all-gt`

```py
#!/usr/bin/env python3
# -*- coding: utf-8 -*-


import sys
import unicodedata
from collections import Counter, OrderedDict


def show_char_frequency(string):
    string = string.replace(" ", "")
    dic = OrderedDict(Counter(string).most_common())
    for char in dic:
        print(f"{char}\t{dic[char]}\t{unicodedata.name(char)}")


def read_file(filename):
    with open(filename, encoding="utf-8", mode="rt") as fd:
        text_lines = fd.read().strip().split("\n")
    return " ".join(text_lines)


def main():
    if len(sys.argv) < 2:
        print(f"USAGE: {sys.argv[0]} <txt_file>")
        return 1

    filename = sys.argv[1]
    string = read_file(filename)
    show_char_frequency(string)
    return 0


if __name__ == "__main__":
    main()
```

count_chars.py
stweil: @cmroughan, are you the author of that code, and would you agree to use it here under the Apache 2.0 License?

cmroughan: I am, and I wouldn't mind its use here under Apache 2.0. I will follow up with a pull request.

stweil:
Commit 29d394b7253c9f933a7fdf57f553d305576f9a5d merged the modifications (based on the original code) which were made in this pull request.

Issue #200, opened 2020-10-11
question: How to Diagnose Overfitting and Underfitting of Tesseract Models?


Shreeshrii:
Is there a way to diagnose Overfitting and Underfitting of Tesseract Models?

@stweil had suggested in a different thread that one should evaluate all models to find the best fit for the eval data. 

Is there a way to extract the checkpoint details of number of iterations and training CER and CER from `lstmeval`  to find the `best` model and graph the results?

wrznr:
@Shreeshrii Not sure about `lsmeval` but the file names of the checkpoints itself encode training CER and number of iterations:
```
$ ls -lrt
total 891036
-rw-rw-r-- 1 kmw kmw 34427398 Sep  2 13:13 boernerianus_checkpoint
-rw-rw-r-- 1 kmw kmw 17212641 Sep  2 13:13 boernerianus_71.96_9294_9300.checkpoint
-rw-rw-r-- 1 kmw kmw 17212665 Sep  2 13:13 boernerianus_68.251_9493_9500.checkpoint
-rw-rw-r-- 1 kmw kmw 17212689 Sep  2 13:13 boernerianus_64.962_9689_9700.checkpoint
-rw-rw-r-- 1 kmw kmw 17212713 Sep  2 13:13 boernerianus_62.498_9889_9900.checkpoint
-rw-rw-r-- 1 kmw kmw 17212737 Sep  2 13:13 boernerianus_61.812_10087_10100.checkpoint
-rw-rw-r-- 1 kmw kmw 17212749 Sep  2 13:13 boernerianus_59.662_10185_10200.checkpoint
-rw-rw-r-- 1 kmw kmw 17212785 Sep  2 13:13 boernerianus_57.284_10471_10500.checkpoint
-rw-rw-r-- 1 kmw kmw 17212821 Sep  2 13:13 boernerianus_54.814_10761_10800.checkpoint
-rw-rw-r-- 1 kmw kmw 17212857 Sep  2 13:13 boernerianus_52.799_11221_11300.checkpoint
-rw-rw-r-- 1 kmw kmw 17212917 Sep  2 13:13 boernerianus_49.958_11685_11800.checkpoint
-rw-rw-r-- 1 kmw kmw 17212941 Sep  2 13:13 boernerianus_45.862_11873_12000.checkpoint
-rw-rw-r-- 1 kmw kmw 17212953 Sep  2 13:13 boernerianus_44.034_11965_12100.checkpoint
[...]

PR #236, opened 2021-03-16
Makefile based plotting

Shreeshrii:


M3ssman:
@Shreeshrii I was trying to get on with the plotting, but I stumbled upon the term "validationlist". 
What does this mean in this context? AFAIU, the plotting requires an already existing *.traineddata-model, which is passed to `lstmeval` Tool. Do I need, besides my previous training data, an additional list of lstmf-files to measure the performance? By now, I do it the hard way, which means put the new model into tessconfigs and do a fresh run with real image data.

Shreeshrii:
Usually for training we use a training list and eval list. You can use a different dataset for validation, if you want.

tfukumori:
Thanks for the great feature! I was able to draw a plot.

But I could not get the result of "Make TSV with Eval CER".
It seems that `Eval Char` is not included in `data/ocrd.log`, so it was not output.

I think there is a problem with my environment or steps, but I would appreciate your advice if you can help me.

# Environment and Steps.
## Environment
```
$ tesseract --version
tesseract 5.0.0-alpha-20210401
 leptonica-1.80.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.2) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
```

## Steps 
I think "validate" should be prepared separately from "eval", but for now, I have set the same thing as "eval" in "validate".

```
wget https://github.com/tesseract-ocr/tessdata_best/raw/master/frk.traineddata -O ~/tessdata_best/frk.traineddata

unzip ocrd-testset.zip -d data/ocrd-ground-truth

nohup make training MODEL_NAME=ocrd \
	START_MODEL=frk \
	TESSDATA=~/tessdata_best \
	MAX_ITERATIONS=10000 > data/ocrd.log

cp data/ocrd/list.eval data/ocrd/list.validate

bash -x plot.sh ocrd validate 10
```

# Result
The logs did not contain any logs about "eval", unlike the logs described in [the Issue](https://github.com/tesseract-ocr/tesstrain/issues/200).
## Log
[ocrd.log.zip](https://github.com/tesseract-ocr/tesstrain/files/6628153/ocrd.log.zip)
## TSV
[ocrd-validate-cer.tsv.zip](https://github.com/tesseract-ocr/tesstrain/files/6628150/ocrd-validate-cer.tsv.zip)
## Plot
ocrd-validate-cer

stweil:
Can we merge this pull request, or would anybody suggest additional changes?

plot.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# `make` needs to be run twice, 
+# first to generate traineddata and then to generate new validate.log files.
+# $1 - MODEL_NAME
+# $2 - VALIDATE_LIST
+# $3 - maximum CER for y axis - adjust based on graph
+
+cd plot
+make MODEL_NAME=$1 VALIDATE_LIST=$2 Y_MAX_CER=$3
+make MODEL_NAME=$1 VALIDATE_LIST=$2 Y_MAX_CER=$3
```
bertsky, line 10: If you need a shell script to run a makefile, then that makefile is poorly written (or documented).

Assuming it is a good choice to have a separate makefile for plotting in the first place, I think the plot/Makefile should just require/assume that the toplevel Makfile's `traineddata` target has already been run.

In the very least, this should read `make traineddata && make plotvalidatecer`, not `make all; make all`.

plot/Makefile
```diff
@@ -0,0 +1,132 @@
...
+FAST_LOG_FILES := $(subst tessdata_fast,tessdata_fast,$(patsubst %.traineddata,%.$(VALIDATE_LIST).log,$(FAST_DATA_FILES)))
+
+#Note: This does not find the new traineddata files from current run. 
+#	Hence `make` needs to be run twice to generate new validate.log files.
+
+$(FAST_LOG_FILES): %.$(VALIDATE_LIST).log: %.traineddata
+	OMP_THREAD_LIMIT=1 time -p lstmeval  \
+		--verbosity=0 \
+		--model $< \
+		--eval_listfile $(TMP_VALIDATE_LIST) 2>&1 | grep "^At iteration" > $@
```
bertsky, line 120: This rule needs still to advertise all its dependencies, esp. `TMP_VALIDATE_LIST` and `FAST_DATA_FILES`.

plot/Makefile
```diff
@@ -0,0 +1,132 @@
...
+
+# Concatenate all validate.log files along with their filenames so as to include iteration number for TSV.
+$(TMP_FAST_LOG): $(FAST_LOG_FILES)
+	@for i in $^; do \
+		echo Filename : "$$i";echo;cat "$$i"; \
```
bertsky, line 125: What are the `Filename` lines used for?

plot/plot-eval-validate-cer.py
```diff
@@ -0,0 +1,141 @@
...
+args = arg_parser.parse_args()
+
+maxticks=5
+maxCER=int(args.ymaxcer) #max y axis to display
+
+ytsvfile = "tmp-" + args.model + "-" + args.validatelist + "-iteration.tsv"
+ctsvfile = "tmp-" + args.model + "-" + args.validatelist + "-checkpoint.tsv"
+etsvfile = "tmp-" + args.model + "-" + args.validatelist + "-eval.tsv"
+vtsvfile = "tmp-" + args.model + "-" + args.validatelist + "-validate.tsv"
+plotfile = "../data/" + args.model + "/plot/" + args.model + "-" + args.validatelist + "-cer.png"
```
bertsky, line 32: Hard-coding the values in Python redundantly instead of passing the fully defined path names from the makefile is really bad style.

plot/Makefile
```diff
@@ -0,0 +1,132 @@
...
+	done > $@
+
+# Combine TSV files with all required CER values, generated from training log and validation logs. Plot.
+$(TSV_VALIDATE_CER): $(TMP_FAST_LOG) $(TMP_100_ITERATIONS) $(TMP_CHECKPOINT) $(TMP_EVAL) $(TMP_VALIDATE) 
+	@cat $(TMP_100_ITERATIONS) $(TMP_CHECKPOINT) $(TMP_EVAL) $(TMP_VALIDATE) > "$@"
+	python plot-eval-validate-cer.py -m $(MODEL_NAME) -v $(VALIDATE_LIST) -y $(Y_MAX_CER)
```
bertsky, line 131: This should really be a separate rule with its own target – the concret plot PNG – and passing all the path names of the necessary input files.

Relying on `Y_MAX_CER` to be defined outside the makefile (instead of, say, a default), is bad style.

plot/Makefile
```diff
@@ -0,0 +1,132 @@
...
+	@rm tmp-$(MODEL_NAME)-$(VALIDATE_LIST)*.*
```
bertsky, line 132: Should be placed into a separate (phony) rule (like `clean`), and instead of redefining the filenames implicitly (`tmp-*`), re-use the actual above definitions.

bertsky (changes requested):
While I do very much welcome your effort to integrate a decent plotting facility, I have strong reservations about this implementation and about the way this PR is set up technically.

Beyond the comments given inline, I see two general questions: 
1. How do the two old Python scripts relate to the single new one? (There seems to be lots of code re-use. It's hard to compare to the previous version this way.)
2. How does this relate to the [bigger problem](https://github.com/tesseract-ocr/tesstrain/issues/261) of CER not being calculated correctly by lstmtraining/lstmeval and checkpoints only being created at arbitrary intervals?

Shreeshrii:
@bertsky

Thank you for your detailed feedback. I agree that this may not be the best
way to implement the plotting facility for tesseract training. The scripts
are what I used at that point of time and it helped me visualise the
training process and results.

This PR can be closed if you or others provide an alternative
implementation.

bertsky:
Thanks for your fast feedback @Shreeshrii. In that case I suggest we do close, and (being very much interested in this feature) I promise I'll revisit as soon as #261 is out of the way. (I will strive for a close integration with lstmtraining to output data files, and with the tesstrain makefile, and then re-use your plotting code.)

Shreeshrii:
@bertsky I am closing this. Hope you will add a better implementation soon.

whisere:
I have the same issue as https://github.com/tesseract-ocr/tesstrain/pull/236#issuecomment-858253025
also reported in https://github.com/Shreeshrii/tesstrain-ben/issues/1

Shreeshrii:
@bertsky is planning to improve the output CER as discussed in above thread and will redo the plotting feature.

I have added some more hacks to the scripts for my own personal use based on the above feedback. I will look to posting them in a repo and post a link here as a workaround till the official update.

whisere:
That sounds great! Thank you!

Shreeshrii:
See https://github.com/Shreeshrii/tess5train-fonts

Compare the plot files for finetuning and replacing a layer and you can see that eval does run for replace a layer training.

https://github.com/Shreeshrii/tess5train-fonts/blob/main/data/engFineTuned/plots/engFineTuned-LOG-2.png

https://github.com/Shreeshrii/tess5train-fonts/blob/main/data/engLayer/plots/engLayer-LOG-2.png

whisere:
Many thanks! I will try that!

whisere:
Just to clarify am I right that it still doesn't plot eval data for fine tuned training (only for replacing a layering training)?

Shreeshrii:
The scripts create a tsv from the log file generated during training process. If tesseract does not run it then the info won't be in the log file and won't get plotted. Also, the eval info does not include the training iteration number, it only has the learning iteration number.

As an alternative I run lstmeval on each of the checkpoints and plot that separately, that is the lstmeval after training. I have also added impact centre's ocreval as well as ISRI evaluation's accuracy info. Plotting of accuracy is not yet implemented.

whisere:
Many thanks for the excellent work! I have tried, but the script doesn't work with the ground truth tif and txt pairs for fine tuning we have... it seems to require model.training_text file and reports:

[01:32:54] INFO - === Starting training for language eng
[01:32:54] INFO - Testing font: Arial Bold
[01:32:54] ERROR - Could not find font named Arial Bold.
Pango suggested font DejaVu Sans Bold.
Please correct --font arg.

[01:32:54] INFO - Program /usr/bin/text2image failed with return code 1. Abort.
[01:32:54] INFO - === Phase I: Generating training images ===
[01:32:54] CRITICAL - Required/expected file 'data/ground-truth/modelname-eval.training_text' does not exist
Makefile:437: recipe for target 'data/gtd/list.eval' failed
make: *** [data/gtd/list.eval] Error 1

with
nohup bash 2-training.sh eng Latin eng modelname FineTune 9999 > data/logs/modelname.LOG &

I will keep investigating.

Shreeshrii:
Yes, as my repo name indicates this is the version for training from fonts and training text. However the plotting part only depends on the log file from training. 

I will upload a different version that works with existing tesstrain makefile.

whisere:
Amazing! I will try when it is available : )

And the lstmeval commands does work and are useful with CER and WER output with the original make output, although not showing in plotting:

nohup lstmeval --model data/modelname.traineddata --eval_listfile data/modelname/list.eval --verbosity 2 > data/modelname-eval-list.log &

nohup lstmeval --model data/eng.traineddata --eval_listfile data/modelname/list.eval --verbosity 2 > data/eng-eval-list.log &

PR #237, opened 2021-03-16
tesstrain.py cleanup

Shreeshrii:
Remove commented out code for legacy engine as well as code not used for lstm training


kba (commented):
AFAICT this removes commented out code, adapts logging and removes dead code. LGTM.

stweil (commented):
I am not sure whether removing the legacy training code is a good idea.

Maybe it would be better to maintain that part as long as the legacy recognition is useful.

bertsky:
> I am not sure whether removing the legacy training code is a good idea.
> 
> Maybe it would be better to maintain that part as long as the legacy recognition is useful.

Good point! Keeping this alive does not cost much, but offers many options (that other engines don't). One scenario is retraining the `osd.traineddata` to include more scripts and improve its discriminative ability (esp. regarding Greek which I find very weak).

PR #238, opened 2021-03-16
Add makefile based training text and font to model scripts

Shreeshrii:
Builds a model using synthetic training data generated from 
- training text
- evaluation text
- fonts

A helper bash script is provided to ensure that all required values are passed to makefile.

The box/tiff pairs are saved for the evaluation data and gt.txt is generated from the box files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@ wackenroder_herzensergiessungen_*.tif
 master.zip
 plot/*.LOG
 plot/ocrd*
+data/
+src/training/__pycache__/
+src/training/tesstrain_utils-save.py
+src/training/tesstrain_utils-singlelines.py
--- a/Makefile-font2model
+++ b/Makefile-font2model
@@ -1,0 +1,227 @@
+export
+
+# Disable built-in suffix rules.
+# This makes starting with a very large number of GT lines much faster.
+.SUFFIXES:
+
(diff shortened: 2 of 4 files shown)


bertsky:
Very interesting!

Could we rework this so that we end up with a minimal set of changes to the base Makefile instead? (That would avoid the need for synchronization of the two variants in the future, plus we could try add the possibility to **mix** synthetic and GT training.)

bertsky:
These stale .github rules are a joke. Auto-close of issues and PRs? In this kind of repo? 

What was the rationale behind this, @wrznr?

wrznr:
The rational was to prevent user questions which require detailed information from the reporter which often never comes from being open issues forever. We may have to adjust the rules for a more realistic behavior.

bertsky:
> The rational was to prevent user questions which require detailed information from the reporter which often never come from being open issues forever. We may have to adjust the rules for a more realistic behavior.

I see. But that problem can be addressed directly/manually IMV: just add a comment with a friendly reminder ("or can we close?"). This can be done when and if maintainers/contributers here do have the time. Right now it's the opposite: we are forced to tend to the machinery within so and so many days, or otherwise things will disappear from the radar. (And assigning labels is also effort, esp. if you don't recall what labels are needed for what kind of promotion.)

PR #249, opened 2021-05-19
Add --vertical_fontlist option to tesstrain.py

nagadomi:
Porting from https://github.com/tesseract-ocr/tesseract/pull/3434 (not merged) .
>This Pull Request adds --vertical_fontlist option to tesstrain.sh to specify a list of fontnames to render vertical text.
>The format for specifying a list of fontnames is the same as for --font_list option.
>If --vertical_fontlist <FONTS> option is specified, it will override the VERTICLA_FONTS variable(defined in language-specific.sh) with the specified list of fontnames.

>In the current version, the VERTICAL_FONTS variable is hardcoded in language-specific.sh. So, when creating training data for vertical text such as Japanese, users need to edit the source code even if they specify a list of fontnames with --fontlist and --font_dir.

### DIFF
--- a/src/training/tesstrain_utils.py
+++ b/src/training/tesstrain_utils.py
@@ -141,6 +141,13 @@ parser.add_argument(
     type=str,
     help="A list of fontnames to train on.",
 )
+parser.add_argument(
+    "--vertical_fontlist",
+    dest="vertical_fonts",
+    nargs="+",
+    type=str,
+    help="A list of fontnames to render vertical text.",
+)
 parser.add_argument("--fonts_dir", help="Path to font files.")
 parser.add_argument("--tmp_dir", help="Path to temporary training directory.")
 parser.add_argument(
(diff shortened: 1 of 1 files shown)


bertsky:
Looks like this was closed accidentally

nagadomi:
For vertical text, it is hard to use without this option.
However, I have developed a more powerful script to replace text2image/tesstrain.py, so I don't use tesstrain.py anymore.

zdenop:
If I understand "Makefile training" correctly, this ["src/tesstrain"](https://github.com/tesseract-ocr/tesstrain/tree/main/src/tesstrain) is not used for training at the moment.
AFAIK these python script are based on old shell training scripts. I suggest to keep them and maybe we can find the way how to integrate it to current training process...

zdenop:
@nagadomi : can you update this PR to recent git code?

stefan6419846:
According to my request in #307, there basically are two ways to train Tesseract inside this repository:

* The Makefile-based approach to train on real data.
* The Python-based approach to train on artificial data, corresponding to the old Bash-based approach.

My suggestion had been to actually document this somewhere as this is not always clear, but due to the harsh stale automation and (at least in the past) rather restricted responses, this has been buried into closed issues.

zdenop:
As I mentioned in other PR - I am interested in python based training as "make training" is difficult run on windows and it requires tool that could be easily replaced by python (unzip, wget, bc ...). I would suggest to merge/rework current open PR and then move forward (review issues/PR marked as "stale"...

zdenop:
committed in https://github.com/tesseract-ocr/tesstrain/commit/2c7c6e8feaf8aa1f2d1750b689fa46473453885e

PR #250, opened 2021-05-20
Fix CRLF issue on Windows

nagadomi:
On Windows, `Path.write_text` converts the newline code(\n) to CRLF, and `lstmtraining` cannot handle CRLF properly, so the `Deserialize header failed` error occurs. https://github.com/tesseract-ocr/tesseract/issues/2456

This pull request changes `tesstrain.py` to use LF newline code when creating `{lang_code}.training_files.txt` (list of lstmf files) even on Windows.

### DIFF
--- a/src/training/tesstrain_utils.py
+++ b/src/training/tesstrain_utils.py
@@ -709,7 +709,8 @@ def make_lstmdata(ctx):
 
     lstm_list = f"{ctx.output_dir}/{ctx.lang_code}.training_files.txt"
     dir_listing = (str(p) for p in path_output.glob(f"{ctx.lang_code}.*.lstmf"))
-    pathlib.Path(lstm_list).write_text("\n".join(dir_listing))
+    with pathlib.Path(lstm_list).open(mode="w", encoding="utf-8", newline="\n") as f:
+        f.write("\n".join(dir_listing))
 
 # make__traineddata() {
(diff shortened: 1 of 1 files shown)


stweil (approved):

stweil:
It would be interesting to know whether Tesseract training is possible on Windows (without WSL which works). There might be more more problems to solve.

nagadomi:
It seems to work with Anaconda PowerShell(Python env on Windows) and Tesseract5 on Windows, although it's as simple as below.

1. Generate training data from Font and Text with `tesstrain.py`
2. Unpack existing LSTM model with `combine_tessdata`
3. Finetune with `lstmtraining`

One thing to be noted:

[tesserocr](https://github.com/sirfz/tesserocr)'s conda package contains `text2image.exe`, `unicharset_extractor.exe`, `dawg2wordlist.exe`, `lstmtraining.exe` and other executable files, which will be installed in `${HOME}\Anaconda3\Library\bin`. It is in a higher priority PATH on the Anaconda Console than the system Tesseract(`C:\Program Files\Tesseract-OCR`).
So, when I run `python src/training/tesstrain.py ...` command, it runs different version of executable file than I intended, some of the command crash.
The solution is to modify the PATH environment variable before running `tesstrain.py`.
```
$env:path="C:\Program Files\Tesseract-OCR;$env:path"
```

PR #256, opened 2021-06-03
avoid destroying Greek lambda

bertsky:
by using a control character (pretty unlikely to be emitted by text2image or anyone else) instead


kba (approved):
LGTM, using an actual letter like `λ` for this is problematic.

wrznr (approved):

Issue #255, opened 2021-06-03
support the cutoff training regime


bertsky:
How about aiding users in [cut-off training](https://tesseract-ocr.github.io/tessdoc/tess4/TrainingTesseract-4.00.html#training-just-a-few-layers), besides finetuning and from scratch?

We already have `START_MODEL` and `NET_SPEC`, so IIUC we would minimally only need to add some additional variable, say `KEEP_LAYERS`, for the `--append_index` parameter. If undefined, everything is kept (i.e. START_MODEL or NET_SPEC is reused as-is, which amounts to finetuning or from-scratch). But if defined, it would add `--append_index $(KEEP_LAYERS)` (and thus keep only that many layers of START_MODEL and then append new layers from NET_SPEC).

Of course, additional facilities like extracting a traineddata's net-spec, or preconfigured KEEP_LAYERS/NET_SPEC settings for typical scenarios could be offered, too.

PR #257, opened 2021-06-03
documentation: make help

bertsky:


kba (approved):
LGTM. Feel free to merge https://github.com/kba/ocrd-train/commit/bc19025816cfdc75e4acd7511457d3988136a75b for minor inconsistent comments and regenerated `make help` in the README.

Makefile
```diff
@@ -371,3 +372,9 @@ clean-output:
 
 # Clean all generated files
 clean: clean-box clean-lstmf clean-output	
+
+# do not search for implicit rules here:
```
stweil, line 376: `Makefile` already uses `.SUFFIXES`, so that looks redundant.

bertsky: > `Makefile` already uses `.SUFFIXES`, so that looks redundant.

It isn't, though. (I've seen it try lexer and VCS rules.)

The manual says .SUFFIXES is for (old-fashioned) suffix rules only (`.c.o` etc), but this protects against builtin (implicit) pattern rules.

Not sure if using `-R` covers that already, though (but the user might not know about that).

stweil: `make -r` should cover that. Maybe it is sufficient to set `MAKEFLAGS += -r` in the `Makefile`.

bertsky: > `make -r` should cover that. Maybe it is sufficient to set `MAKEFLAGS += -r` in the `Makefile`.

Yes, that does work. (Interestingly, although `make -R` is stronger than `make -r`, `MAKEFLAGS += -R` does _not_ imply `MAKEFLAGS += -rR`. Probably a bug.)  

I'll change the PR accordingly.

bertsky:
> LGTM. Feel free to merge [kba@bc19025](https://github.com/kba/ocrd-train/commit/bc19025816cfdc75e4acd7511457d3988136a75b) for minor inconsistent comments and regenerated `make help` in the README.

Sure! (I assume you produced that using your makefile parser, so consistency is enforced? I wonder what happened to FINETUNE_TYPE ...) @stweil?

stweil:
`FINETUNE_TYPE` was added in commit 0f95a1a7cad11e62df7079fe10b64090ad6a8025, but only to `README.md`. Related changes for `Makefile` never got merged.

bertsky:
> `FINETUNE_TYPE` was added in commit [0f95a1a](https://github.com/tesseract-ocr/tesstrain/commit/0f95a1a7cad11e62df7079fe10b64090ad6a8025), but only to `README.md`. Related changes for `Makefile` never got merged.

Too bad! That's exactly what I was asking for in #255! It seems to have been removed in #134 simply because of the suffix in the model name. IMHO that's not a good reason to remove the option for additional training regimes completely. One could have just restored the model name (e.g. by only reverting c0dd8c9cbbd9611c0dccc4a20019bcc0334a9acf). But @Shreeshrii has also brought other ideas to the table (and removed again), e.g. 85b5ae8ef0a1e5388b81e9bc1af755081f420e92. Let's discuss that in #255!

bertsky:
> Feel free to merge [kba@bc19025](https://github.com/kba/ocrd-train/commit/bc19025816cfdc75e4acd7511457d3988136a75b) for minor inconsistent comments and regenerated `make help` in the README.

@kba sry, you'll have to make a new PR now, this has already been merged.

PR #264, opened 2021-06-15
correct makefile

akash-akya:
Looks like this is added by mistake recently


stweil (approved):
Thank you.

stweil:
> Looks like this is added by mistake recently

Yes, it was added accidentally in commit ee0a8de9.

Issue #308, opened 2022-05-20
Python package for tesstrain.py


stefan6419846:
The Python approach in *src/training* requires additional effort to be used from within own Python code, as there is no PyPI package for it.

It would be great to make this code a real Python package, which can be used as a library. This requires some refactoring to work. The existing `tesstrain.py` script can be kept as an entry point for direct usage.

Issue #307, opened 2022-05-19
Migrating from tesstrain.sh


stefan6419846:
I have used the `tesstrain.sh` approach (including `tesstrain_utils.sh` and `language-specific.sh`) for fine tuning an existing model for a specific font in the past. As this is deprecated with the corresponding Bash scripts having been removed from the *tesseract* repository, I wanted to try the new approach which redirects to this repository.

Looking at this repository, it seems to provide four different types of training support:

* A Makefile-based approach with short documentation inside the README file.
* Some Python scripts in the root directory, which seem to be partially used by the Makefile.
* Some plotting functionality in the *plot* directory.
* A Python-based replacement for the old Bash scripts in *src/training*.

Given this situation, how do the four different "types" interact with each other? What is the correct approach to use for training, given that I want to avoid another deprecation after a short time?

Background: As a Python developer I considered using the approach from *src/training*, requiring less migration effort as well. But as this does not seem to be documented in the README, I am not sure whether this makes sense.

As an additional question: The module in *src/training* does not seem to be available as a regular Python package on PyPI, although it seems like it could be. Are there any plans to convert this to a library (leaving `tesstrain` as an entry point for standalone execution), making it easier to use this in own code without maintaining local copies?

stefan6419846:
Thanks for the explanations about the Makefile (and the files inside the root directory) being intended for "real-life" training, while *src/tesstrain* keeping the artificial approach alive.

In my opinion (mostly being a Tesseract user instead of a Tesseract developer) these aspects should be represented in the directory structure as well. This might be achieved by moving all the sources for "real-life" training into an own subdirectory and maybe improving the naming for the artificial approach directory. Then each directory could have a dedicated README, while the global README provides some basic explanations.

Regarding the Python package I am going to open a new issue.

stefan6419846:
Is there any interest in actually cleaning up the directory structure and improving the corresponding documentation? If yes, does it make sense to track it in this issue, or should this rather be a new one?

PR #309, opened 2022-05-24
Migrate Python code to a dedicated package

stefan6419846:
This is my attempt to migrate the existing code for working with artificial training data to a dedicated Python package, as proposed in #308 and #307. This includes some additional refactoring to the module structure to better encapsulate specific functionality.

I have used version number 0.1 for now, although I am up to changing this.

When migrating, I had two parameters which I am not sure about:

* `overwrite`, defaulting to `False`, does not seem to be used at all.
* It has not been clear enough for me what `extract_font_properties` really means and therefore misses documentation. `text2image --help` did not really help me in this case as well:

  >   --only_extract_font_properties  Assumes that the input file contains a list of ngrams. Renders each ngram, extracts spacing properties and records them in output_base/[font_name].fontinfo file.  (type:bool default:false)

  What would be an appropriate documentation of the parameter?

If there is anything unclear or you want to see anything changed about this, feel free to ask or report.


Shreeshrii:
Thanks. A dedicated package for training from fonts is a good idea.

You may want to look at the original bash scripts that were sought to be replicated in these python scripts in older versions eg. https://github.com/tesseract-ocr/tesseract/tree/4.0/src/training

'overwrite' if I recall correctly was used for the legacy training offered in the bash scripts.

src/README.md
```diff
@@ -0,0 +1,20 @@
+# tesstrain.py
+
+Utilities for working with Tesseract 4 using artificial training data.
+
+## Install
+
+This package requires the Tesseract training tools to be available on your system.
+
+To install the PIP package, either use `pip install tesstrain` or `python setup.py install`. At least Python 3.6 is required.
```
stweil, line 9:
```suggestion
To install the PIP package, either use `pip install tesstrain` or `python setup.py install`. At least Python 3.7 is required.
```

stweil: Python 3.6 is an old version which is no longer supported, so there is no need to support it here.

src/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    packages=setuptools.find_packages(),
+    license='Apache Software License 2.0',
+    author='Tesseract contributors',
+    classifiers=[
+        'Development Status :: 5 - Production/Stable',
+        'Intended Audience :: Developers',
+        'Topic :: Scientific/Engineering :: Image Recognition',
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.6',
```
stweil, line 23:
```suggestion
```

src/setup.py
```diff
@@ -0,0 +1,39 @@
...
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+    ],
+    keywords='Tesseract,tesseract-ocr,OCR,optical character recognition',
+
+    python_requires='>=3.6',
```
stweil, line 30:
```suggestion
    python_requires='>=3.7',
```

stweil:
Font properties (bold, italic, ...) are also from the legacy training and still unsupported with the LSTM recognizer. That's one of the reasons why there remains a certain need for legacy models. The old `tesstrain.sh` supported training of legacy models, and I think that it would be good to support it in the Python code, too. That should be done separately, not in this pull request here, but maybe you can keep the corresponding parameters with appropriate TODO comments.

src/README.md
stefan6419846: You are right about this, although the code should run fine on Python 3.6 for now and therefore it should not hurt. To be honest, I actually used Python 3.6 for the migration as this our company did not yet migrate from it.

stefan6419846:
@stweil Do you mean that we should keep the existing parameters for now when you are talking about the legacy support? Or does this refer to the `linedata` parameter only?

src/README.md
stweil, line 9:
```suggestion
To install the PIP package, either use `pip install tesstrain` or `python setup.py install`.
It works with Python 3.6, but a supported Python version (at least 3.7) is recommended.
```

stweil:
> @stweil Do you mean that we should keep the existing parameters for now when you are talking about the legacy support? Or does this refer to the `linedata` parameter only?

I'd keep all existing parameters for now (with comments).

stefan6419846:
I have updated the requirements inside the README and fixed the parameters for `tesstrain.wrapper.run()`.

stefan6419846:
Is there anything to be changed here to get this PR merged?

.gitignore
```diff
@@ -21,3 +21,4 @@ master.zip
 main.zip
 plot/*.LOG
 plot/ocrd*
+.idea/
```
stweil, line 24: Personally I prefer a minimalistic `.gitignore` which only contains entries related to the project.

Ignoring files from individual editors and development tools should be ignores in centralized personal settings.

src/.gitignore
```diff
@@ -0,0 +1,160 @@
+# Byte-compiled / optimized / DLL files
```
stweil, line 1: This whole file looks like a copy from somewhere else. Which parts are really required? Why can those parts not be included in the root `.gitignore`?

kba: I think this is the [standard GitHub Python .gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore)

.gitignore
stefan6419846: I have removed the offending entry. In fact I have not heard of global *.gitignore* configurations before, as my projects usually have a different directory structure.

src/.gitignore
stefan6419846: Yes, this is the default Python *.gitignore* template which usually is recommended for Python projects.

Only keeping the relevant entries might be an option, mainly concerning specific libraries which are currently unused. Nevertheless, I usually prefer to keep the default *.gitignore* file as it usually is some type of common sense.

I did not add this to the root *.gitignore* file to avoid unintended side effects for now, as the whole repository consists of two more or less independent parts, while not being clearly separated for now (see #307).

stefan6419846: @stweil I have reduced the `.gitignore` to the entries which actually make sense in this use case (at least from my point of view).

Do you still want me to migrate the corresponding entries to the root `.gitignore` file?

stefan6419846:
Any further update on this?

zdenop:
Can this be merged?

stefan6419846:
At least from my side, yes. @stweil had some objections about the `.gitignore` file, but I did not yet hear back after my latest request for further changes which might be required.

stweil:
I'd remove any reference to Python 3.6 (see my two comments). `.gitignore` still contains lots of entries which are not strictly necessary, but that is not critical for merging.

stefan6419846:
I have removed the references to Python 3.6 and updated the README to make clear that Tesseract version 5 is supported as well.

stweil (approved):
Let's try the new code. Thank you @stefan6419846 for your patience.

zdenop:
@stweil: what about tagging the previous code/commit as version 1.0? 
So we can maybe do more reorganization of code without breaking somebody's workflow.
I am interested to make training on windows just by using python. If python is required then [Auxiliaries](https://github.com/tesseract-ocr/tesstrain#auxiliaries) are really not needed.

PR #343, opened 2023-05-02
Make last checkpoint also precious

stweil:
The last checkpoint is useful for resuming a training session that has been interrupted for whatever reason, so don't delete it.


zdenop:
.. #318 should be closed?

stweil:
Thank you for the hint. I missed that PR, but merged it now. So my PR is no longer needed.

PR #364, opened 2023-12-13
Remove build rules for Tesseract and Leptonica

stweil:
Recent releases of Tesseract and Leptonica are available for all major platforms (Linux, MacOS, Windows), so local build rules are more confusing than helpful.

Update also the links to the online documentation.


zdenop:
thanks

PR #374, opened 2024-03-01
fix make detection of version

bertsky:
Commit 25b8508f06365de0fe004fa09871e9bda1b56694 introduced a bug which now prevents `charfreq` from working.

Also, it was not necessary to do the make version checks as a dependency with shell invocation; rather, it's simply a matter of using make's conditionals and control functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,17 +163,16 @@ help: default
 
 # END-EVAL
 
-default:
 ifeq (4.2, $(firstword $(sort $(MAKE_VERSION) 4.2)))
-   # stuff that requires make-3.81 or higher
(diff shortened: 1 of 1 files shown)


stweil (approved):
Thank you.

stweil:
@bertsky, I am curious: does `charfreq` work for you at all? It fails for me with combined characters like aͤ.

bertsky:
@stweil sure, for combining sequences it did not work, but the multibyte problem points to bugs in the MacOS installation of libc localization facility.

I'll prepare a PR covering combining sequences properly.

PR #378, opened 2024-03-08
contributing count_chars.py utility

cmroughan:
As discussed in https://github.com/tesseract-ocr/tesstrain/pull/235#discussion_r1517037975_ .

Would a pull request from myself be sufficient here, or would it be preferable to add some language about this contribution being under the Apache-2.0 license?

### DIFF
--- a/count_chars.py
+++ b/count_chars.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+
+import unicodedata
+import sys, getopt
+
+def main(argv):
+    txt_file = ''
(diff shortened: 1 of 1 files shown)


stweil:
Thank you! It would be great if you could add a statement about the license either as commit comment or in the header of the code.

cmroughan:
Sounds good, I've updated the language to make the licensing here more explicit.

stweil (approved):
Thanks a lot!

PR #379, opened 2024-03-10
Use python script for generating list.eval and list.train.

zdenop:
This decreases dependency on `bc` program and solves the problem with EOL on Windows.

### DIFF
--- a/generate_eval_train.py
+++ b/generate_eval_train.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import pathlib
+import sys
+
+
+def split_file(input_file, ratio):
+    """
+    Splits a text file into list.train and  list.eval with lines ratio.
+    """
+    if not isinstance(input_file, pathlib.Path):
+        input_file = pathlib.Path(input_file)
+    if not input_file.exists():
+        print(f"'{input_file}' not exists!")
+        return False
+    lines = input_file.read_text().splitlines()
+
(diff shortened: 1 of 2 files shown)


generate_eval_train.py
```diff
@@ -0,0 +1,39 @@
...
+    split_point = int(ratio * len(lines))
+    output_dir = input_file.resolve().parent
+    train_list = pathlib.Path(output_dir, 'list.train')
+    eval_list = pathlib.Path(output_dir, 'list.eval')
+
+    with open(train_list, 'w', newline='\n') as f1, open(
+        eval_list, 'w', newline='\n'
+    ) as f2:
+        f1.writelines(lines[:split_point])
+        f2.writelines(lines[split_point:])
```
stweil, line 28:
```suggestion
    with open(train_list, 'w', newline='\n') as f:
        f.writelines(lines[:split_point])
    with open(eval_list, 'w', newline='\n') as f:
        f.writelines(lines[split_point:])
```

stweil: I think it looks simpler like that. Is there a reason to implement it differently?

stweil (changes requested):

generate_eval_train.py
zdenop: As far as I see modern examples and tutorials on the internet: if you open more files with `with` they are always combined into one as in my PR.
Also, python linters do not like single-letter (`f`) variables and for readability, I suggest using different names for eval and train files.

stweil:
Thank you.

